PR #2, opened 2015-08-13
fix issue #1 -- exclude srcsrv files from nuget

### LINKED ISSUE #1: In the Prajna nuget package, tools\Client should not contain the .srcsrv files
In the Prajna nuget package, tools\Client should not contain the .srcsrv files. These files were generated as by product of SourceLink, should not be included in the package.

image

zhuweirong:
Fixes #1


zhuweirong:
Looks OK.

PR #13, opened 2015-08-15
Change DSet.Fold to use state separately per Parition. 

jinlmsft:
The pull request address issue #12.

Change DSet.Fold to use state separately deserialized per partition. A FoldWithCommonStatePerNode is retained for internal implementation that takes advantage of common state per node.

### DIFF
--- a/src/CoreLib/DKV.fs
+++ b/src/CoreLib/DKV.fs
@@ -85,7 +85,7 @@ type DKV<'K, 'V when 'K : equality> =
             else
                 null
         let d = x |> DSet.cacheInMemory |> DSet.map (fun (k, v) -> k )
-        let set = d.Fold(foldFunc, aggrFunc, null)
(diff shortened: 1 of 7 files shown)


zhuweirong:
Should not include the change to 
- docs/content/index.fsx
- docs/tools/templates/template.cshtml

Should include a unit test that verifies the behavior of fold.  For example, we can have a test that has 4 partitions on two nodes. The elements in the partitions all have value 0. Then we have "State" as 1.  Both folder and aggregator is "+". Then the expected result should be "4 + 2 = 6".  Before the fix, the result would be "2 + 2 = 4".

zhuweirong:
Please try to squash all commits related to a single issue into a single commit. See: https://github.com/ginatrapani/todo.txt-android/wiki/Squash-All-Commits-Related-to-a-Single-Issue-into-a-Single-Commit

wezhu:
test

PR #14, opened 2015-08-17
Adding map-reduce version of Naïve Bayes, null check for Fold version

BrunoBozza:
For the Fold version, for now I am passing null for the zero object, which will work with either Fold semantics. Once Jin's commit is in we can update the example to show both semantics.

### DIFF
--- a/samples/examples/FSharpExamples/NaiveBayes.fs
+++ b/samples/examples/FSharpExamples/NaiveBayes.fs
@@ -31,10 +31,24 @@ open Prajna.Api.FSharp
 
 open Prajna.Examples.Common
 
+/// A Counts object holds word counts for a single class
+/// A counts: Count[] holds counts for class k in in counts.[k]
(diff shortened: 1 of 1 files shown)


samples/examples/FSharpExamples/NaiveBayes.fs
```diff
@@ -1,4 +1,4 @@
-(*---------------------------------------------------------------------------
+﻿(*---------------------------------------------------------------------------
```
zhuweirong: Maybe adding a little bit more comments to the code?  Unlike wordcount or matchword that is very trivial, this one may not be that apparent to users without background.  Some comments on the over idea may help.

zhuweirong:
Maybe adding a little bit more comments to the code? Unlike wordcount or matchword that is very trivial, this one may not be that apparent to users without background. Some comments on the over idea may help.

BrunoBozza:
Done.

PR #15, opened 2015-08-18
Measure test time, kill hanging nunit-agent.exe, and fix data drive selection when there's only a single drive

zhuweirong:
1. Kill hanging nunit-agent.exe if any when unit test completes
2. Measure test time
3. When there's only when drive, do not exclude it for data drive. Also handle mono differently.


hongzhili:
It looks good. Please merge this pull request. I'll fix a few minor issues after it is merged.

PR #35, opened 2015-09-03
For C#, add Linq API mappings in namespace Prajna.Api.CSharp.Linq

zhuweirong:
For C#, add Linq API mappings in namespace Prajna.Api.CSharp.Linq
- Only some of the common Linq APIs are mapped
- it's a conceptual map, not a strict 1-1 map
- Some mappings like "SelectMany -> Mapi", the parameters for SelectMany is different from Linq API, since we have the index of partition and elements in partition.

### DIFF
--- a/src/CoreLib/DSetCSharpLinq.fs
+++ b/src/CoreLib/DSetCSharpLinq.fs
@@ -1,0 +1,99 @@
+﻿(*---------------------------------------------------------------------------
+    Copyright 2013 Microsoft
+
+    Licensed under the Apache License, Version 2.0 (the "License");
+    you may not use this file except in compliance with the License.
+    You may obtain a copy of the License at
+
+    http://www.apache.org/licenses/LICENSE-2.0
+
+    Unless required by applicable law or agreed to in writing, software
+    distributed under the License is distributed on an "AS IS" BASIS,
+    WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+    See the License for the specific language governing permissions and
+    limitations under the License.                                                      
+
+    File: 
+        DSetCSharpLinq.fs
+  
+    Description: 
+        CSharp Linq APIs for DSet<'U>
+
+    Author:																	
+        Jin Li, Principal Researcher
+        Microsoft Research, One Microsoft Way
+        Email: jinl at microsoft dot com
+    Date:
+        Sept. 2015
+    
+ ---------------------------------------------------------------------------*)
+namespace Prajna.Api.CSharp.Linq
+
+open System
+open System.Collections.Generic
+open System.Runtime.CompilerServices
+
+open Prajna.Api.CSharp
+
+[<Extension>]
+type DSetLinqExtensions =
+
+    /// Creates a new dataset whose elements are the results of applying the given function to each of the elements 
+    /// of the dataset. 
+    [<Extension>]
+    static member Select (dset : DSet<'U>, func : Func<'U, 'U1>) =
+        dset.Map(func)
+
+    /// Creates a new dataset whose elements are the results of applying the given function to each of the elements 
+    /// of the dataset. The first integer index passed to the function indicates the partition, and the second integer
+    /// passed to the function index (from 0) element within the partition
+    [<Extension>]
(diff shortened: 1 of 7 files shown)


src/CoreLib/DSetCSharpLinq.fs
```diff
@@ -0,0 +1,104 @@
...
+        DSetLinqExtensions.GroupBy(dset, keySelector).Map(fun (k, grp) -> k, grp |> Seq.map(fun v -> elementSelector.Invoke(v))).Map(fun (k, grp) -> resultSelector.Invoke(k, grp) )
+
+    /// Returns the dataset as an IEnumerable<'U>
+    [<Extension>]
+    static member AsEnumerable(dset : DSet<'U>) = 
+        dset.ToIEnumerable()
+
+    /// Produces the set union of two datasets
+    [<Extension>]
+    static member Union(dset1 : DSet<'U>, dset2 : DSet<'U> ) =
```
zhuweirong: Note: Linq's Union uses default equality comparer for type 'U. 

For Prajna's, as it is now, Union merges all records without comparison to remove duplicates.

src/CoreLib/DSetCSharpLinq.fs
```diff
@@ -0,0 +1,104 @@
...
+        dset1.Union(seq { yield dset2 })
```
zhuweirong: Note: Linq's Union uses default equality comparer for type 'U.

For Prajna's, as it is now, Union merges all records without comparison to remove duplicates. We may want to revisit the semantics of Union.

zhuweirong: Note: Linq's Union uses default equality comparer for type 'U.

For Prajna's, as it is now, Union merges all records without comparison to remove duplicates. We may want to revisit the semantics of Union.

BrunoBozza: seq { yield dset2 } is the same as Seq.singleton dset2.

src/CoreLib/DSetCSharpLinq.fs
```diff
@@ -0,0 +1,104 @@
...
+
+    Author:																	
+        Jin Li, Principal Researcher
+        Microsoft Research, One Microsoft Way
+        Email: jinl at microsoft dot com
+    Date:
+        Sept. 2015
+    
+ ---------------------------------------------------------------------------*)
+namespace Prajna.Api.CSharp.Linq
```
BrunoBozza, line 30: Do we need "CSharp" in the namespace? This makes it seem that CSharp is non-native for Prajna. And "Api" is kind of implied. This will be a core, popular API, so it would be nice to have a short name. Why not "Prajna.Linq"?

src/CoreLib/DSetCSharpLinq.fs
BrunoBozza: You mean revisit the semantics of Union in Prajna, right? The C# API should have semantics compatible with Linq.

src/CoreLib/DSetCSharpLinq.fs
zhuweirong, line 30: So we have APIs only for FSharp, and have APIs that is for general .Net (thus can be used by both F#, C#, or VB).  However, the latter is assumed to be used mainly by C# user. So currently the structure is:
- Prajna.Core:  runtime APIs (e.g. Cluster) that can be used from both C#/F#
- Prajna.Api.FSharp: F# only APIs
- Prajna.Api.CSharp: C# APIs, (but in fact usable in any .Net language)
  Prajna.Tools is structured in a similar way. I think we can further discuss the choice here. But this pull request would not address it, since it's a much larger scope of decision.

src/CoreLib/DSetCSharpLinq.fs
zhuweirong: Correct! So if we want to be sure have the same semantics of Linq. I will remove this Union API mapping for now.

BrunoBozza: You could call it "UnionDistinct" or something like that.

src/CoreLib/DSetCSharpLinq.fs
BrunoBozza, line 30: Agreed.

PR #38, opened 2015-09-11
New Prajna serializer, initial pull request

BrunoBozza:
The initial serializer implementation was designed as a _reference_ but is already 2 to 5 times faster on large object graphs. In flat graphs (i.e.: big arrays of primitive types) it's about the same. Final optimized implementation should be even faster as I enable memory block copy.

All Prajna tests are passing and I have added 21 new ones for the serializer functionality, but if you think your private code is breaking because of this, you can turn it off by setting GenericSerialization.DefaultFormatterGuid to GenericSerialization.BinaryFormatterGuid. If you indeed find that this was the problem, it's a bug and I would like know. I am setting PrajnaFormatter to be the default - because it's faster already, and so we can find any bugs.

Things to do, coming on next pull request:
- Supporting 2D and higher rank arrays.
- Re-enabling block copy

For future commits:
- Using CompilerServices.RuntimeHelper for fast, reference-based hash codes
- Code generation instead of Reflection (yes, even using reflection this is much faster than the .net  BinarySerializer)


zhuweirong:
For "Using CompilerServices.RuntimeHelper for fast, reference-based hash codes", please take a look a helper function we already have in Utils.fs -- called Utils.GetHashCode.

src/tools/tools/serialize.fs
```diff
@@ -17,9 +17,20 @@
...
 namespace Prajna.Tools
 open System
+open System.Reflection
+open System.Runtime.InteropServices
+open System.Runtime.Serialization
+open System.Collections.Generic
 open System.IO
 open System.Runtime.Serialization.Formatters.Binary
 
+type ReferenceType =
```
zhuweirong: should it be "internal"?  Add some comments to explain each field?

src/tools/tools/serialize.fs
```diff
@@ -153,3 +164,519 @@ module internal Serialize =
...
+    let CSharpFuncFields = BindingFlags.Instance ||| BindingFlags.Public ||| BindingFlags.NonPublic
+
+    let rec isBlittable (t: Type) = 
+//        let rec fieldsAreBlittable = memo <| fun (t: Type) -> t.GetFields(AllInstanceFields) |> Seq.forall (fun field -> isBlittable field.FieldType)
+        t.IsPrimitive //|| (t.IsValueType && fieldsAreBlittable t)
+
+    let theConverter = FormatterConverter()
+    let theContext = StreamingContext(StreamingContextStates.Remoting)
+
+type SerializationCallback = obj -> StreamingContext -> unit
```
zhuweirong: internal?

src/tools/tools/serialize.fs
```diff
@@ -153,3 +164,519 @@ module internal Serialize =
             assert(false)
             Unchecked.defaultof<'T>
 
+    let inline memo (f: 'a -> 'b) =
+        let cache = new Dictionary<'a, 'b>()
```
zhuweirong, line 180: So it's not thread-safe?

src/tools/tools/serialize.fs
BrunoBozza: Yep. Thanks.

src/tools/tools/serialize.fs
BrunoBozza: Yep.

src/tools/tools/serialize.fs
BrunoBozza, line 180: No, not thread safe. Serialization is inherently sequential, and making this thread safe would have a cost, both in performance and cognitive, which we wouldn't be using. That's why I'm keeping it in module Serialize, not Utils.

BrunoBozza:
RuntimeHelpers.GetHashCode would be used only with reference types, so the typedefof<'T> and IsValueType are not needed.

BrunoBozza:
Oh... turns out I am already using RuntimeHelpers.GetHashCode... it's been a while. I wonder if they are guaranteed to be unique like in Java, so we can use a Dictionary<int,int> for serializing, instead of the current Dictionary<obj, int>.

samples/examples/FSharpExamples/NaiveBayes.fs
```diff
@@ -143,92 +169,144 @@ type NaiveBayes() =
...
-            DSet<string>(Name = name, Cluster = cluster)
-            |> DSet.distributeN 8 trainSet
-            |> DSet.choose chooseLine
-            |> DSet.fold (addWords numClasses) addCounts null
-        printfn "DSet train took: %A" (sw.Stop(); sw.Elapsed)
 
-        sw.Restart()
-        let mapReduceCounts = naiveBayesMapReduce name cluster trainSet
-        printfn "MapReduce train took: %A" (sw.Stop(); sw.Elapsed)
+        let foldPrajna, mapReducePrajna = trainAndTime cluster name "Prajna formatter" trainSet
```
zhuweirong, line 180: In general, we avoid using "Prajna" in type/member/value names (especially for public visible ones).  We want to only use it in namespace and constants. So in the future, if a rename is needed, it won't be too many changes. These are not public visible, so it should be OK. But as a general practice ....

PR #41, opened 2015-09-14
Merge with new Memory Stream implementation

sanjeevm0:


sanjeevm0:
One more file needed to be added

PR #42, opened 2015-09-14
Merge new Memory Stream implementation

sanjeevm0:
There are considerable number of changes in this request.  Some reorganization of code is also there, and there were some conflicts, including some when merging with Bruno's latest changes.  If you see any issues, please let me know.  I have several things which I will look at, but wanted to merge, otherwise there is a lot of effort in continuing to have to merge large number of changes.
1. I have still to move the array serialization used for the sort benchmark to custom serialization
2. I will modify StreamBase and its derivatives to implement IDisposable instead of ref counting and use replication instead of assignment (overlap with Jin's current work likely)
3. The refactoring overlaps with Bruno's refactoring of MemStream, may need second look.
4. For some reason, using our threadpool for networking code causes UT hangs on Azure VM.  Using system threadpool works fine - so I changed networking code to use system threadpool.  This is likely due to some remaining bugs in threadpool which need to be fixed.  It is only a two line change to switch back.


zhuweirong:
This pull requests brings in many commits that will go into master's commit history. Some of them are probably not in working condition.  Could you please consider squash the commits into one or a few meaningful ones for the commit history?

Prajna.sln
```diff
@@ -120,10 +120,6 @@ Project("{F2A71F9B-5D33-465A-A702-920D77279786}") = "Common", "samples\examples\
 EndProject
 Project("{FAE04EC0-301F-11D3-BF4B-00C04F79EFBC}") = "CSharpExamples", "samples\examples\CSharpExamples\CSharpExamples.csproj", "{A68BD032-F86B-49B6-B7ED-7300DB353705}"
 EndProject
-Project("{F2A71F9B-5D33-465A-A702-920D77279786}") = "ServiceEndpoint", "src\ServiceLib\ServiceEndpoint\ServiceEndpoint.fsproj", "{45959765-A1FA-4277-BC68-024CAD648546}"
-EndProject
-Project("{F2A71F9B-5D33-465A-A702-920D77279786}") = "Gateway", "src\ServiceLib\Gateway\Gateway.fsproj", "{C588134A-466C-4453-89AB-7DD7CEC58F06}"
-EndProject
```
zhuweirong: Are these two projects removed accidentally?

src/tools/tools/file.fs
```diff
@@ -228,17 +228,17 @@ module  FileTools =
...
+            Logger.LogF(LogLevel.MediumVerbose,  fun _ -> sprintf "WriteBytesToFileConcurrentPCompare: create and write file '%s'" filename)
+            try 
+                use fileStream = CreateFileStreamForWrite( filename )
+                fileStream.Write( bytes, offset, len )
+                fileStream.Close() 
+            with 
+            | e -> 
+                message <- e.Message
+                bExist <- true
+            //Logger.LogF(LogLevel.MediumVerbose,  fun _ -> sprintf "WriteBytesToFileConcurrentPCompare: complete create and write file '%s' (exists = %b)" filename bExist)
```
zhuweirong: why this line was commented out? It was added intentionally to help track down an issue.

src/tools/tools/file.fs
```diff
@@ -309,7 +309,7 @@ module  FileTools =
                                 nVerified <- -1 
                             else
                                 Threading.Thread.Sleep( 5 )
-            Logger.LogF(LogLevel.MediumVerbose,  fun _ -> sprintf "WriteBytesToFileConcurrentPCompare: done verify file '%s' (verified = %d), detail: %s" filename nVerified message)
+            //Logger.LogF(LogLevel.MediumVerbose,  fun _ -> sprintf "WriteBytesToFileConcurrentPCompare: done verify file '%s' (verified = %d), detail: %s" filename nVerified message)
```
zhuweirong: why this line was commented out? It was added intentionally to help track down an issue.

src/CoreLib/DSeq.fs
```diff
@@ -875,20 +897,25 @@ and [<AllowNullLiteral>]
...
-                        //#tag: disable flowcontrol
-                        // JinL: 06/19/2015, leave flow control to the network layer. 
-                        let bCansend = peerQueue.CanSend && 
-                                         ( 
-                                            (int64(x.SendingQueueLimit) > peerQueue.UnProcessedCmdInBytes + ms.Length)  ||
-                                            ( peerQueue.UnProcessedCmdInBytes = 0L && (int64(x.SendingQueueLimit) < ms.Length  )  ) // to send huge command
-                                         ) 
-                        //let bCansend = peerQueue.CanSend 
+//                    if Interlocked.CompareExchange (peerQueue.flowcontrol_lock,1,0) = 0 then
+////                        #tag: disable flowcontrol
```
zhuweirong, line 901: No need to take the lock any more?

src/CoreLib/DSetGenerics.fs
```diff
@@ -626,6 +628,26 @@ type DSet<'U> () =
...
+        x.NumPartitions <- x.Cluster.NumNodes * x.NumParallelExecution
+        x.NumReplications <- 1
+        x.Dependency <- Source
+        x.Function <- Function.Init<'U>( initFunc, partitionSizeFunc x.NumPartitions ) 
+        // Trigger of setting the serialization limit parameter within functions. 
+        // Automatic set serialization limit
+        // x.SerializationLimit <- x.SerializationLimit
+        x
+
+    /// <summary>
```
zhuweirong: For introducing a new DSet API here, there are a few more places to change
1. Add a corresponding static member in this file
2. Add the corresponding C# API in DSetCSharp.fs

src/CoreLib/GV.fs
```diff
@@ -1190,7 +1194,7 @@ type internal GVSerialize<'K>( ) =
     inherit GVSerialize( 
         let wrapperFunc (ms:MemStream) (O1:Object) =
             let state1 = O1 :?> 'K
-            ms.SerializeFrom( state1 )
+            Strm.SerializeFrom( ms, state1 )
```
zhuweirong: Is "Strm" a type name? What does it stands for?  In general, .Net framework guideline does not prefer using abbreviations for naming unless it's very well known (for example IO).   See:  https://msdn.microsoft.com/en-us/library/141e06ef(v=vs.71).aspx

src/CoreLib/deploy.fs
```diff
@@ -217,10 +219,10 @@ type DeploymentSettings() =
     /// ReSerialization For Cache
     static member val internal DefaultCacheSerializationLimit = 1 with get, set
     /// ReSerialization For Cache
-    static member val internal DefaultIOMaxQueue = 10 with get, set
+    static member val internal DefaultIOMaxQueue = 100 with get, set
```
zhuweirong, line 222: Is this a good default for general use or it's optimized towards sort benchmark?

src/CoreLib/deploy.fs
```diff
@@ -256,9 +258,27 @@ type DeploymentSettings() =
     /// Maximum combined buffer devoted to sending queue (in bytes) 
     static member val MaxSendingQueueLimit = 1024 * 1024 * 50 with get, set
     /// Maximum network stack memory (total across all connections), 0 means unbounded (in bytes)
-    static member val MaxNetworkStackMemory = 1024 * 1024 * 1024
+    static member val MaxNetworkStackMemory = 1024 * 1024 * 1024 * 4 with get, set
```
zhuweirong, line 261: Is this a good default for general use or it's optimized towards sort benchmark?

src/tools/tools/configuration.fs
```diff
@@ -244,7 +244,7 @@ module internal ConfigurationUtils =
                 )
 
     // Unpack asm binding information
-    let UnpackAsmBinding (ms : MemStream) =
+    let UnpackAsmBinding (ms : StreamBase<byte>) =
```
zhuweirong: How about the corresponding PackAsmBinding function, does it also need to change to use StreamBase<byte> instead of MemStream?

jinlmsft: Perhaps PackAsmBinding  and UnpackAsmBinding can consider to use (ms:Stream)?

src/tools/tools/file.fs
zhuweirong: I guess it's because there're compilation error about line 241 with VS2013?  We didn't get it in other places, since we have switched to VS2015.  It seems the new compiler relaxed the capture requirements of closure.  We probably should make it correct for VS2013 too.

samples/DistributedSort/DistributedSort.fs
```diff
@@ -207,27 +213,34 @@ let main orgargs =
...
+        let dkvGen = startDSet.InitN(initVectorFunc, partitionSizeFunc)
+        //let dkvGen = startDSet |> DSet.init initVectorFunc partitionSizeFunc 
         //let dkvGenRAM = dkvGen |> DSet.cacheInRAM
         let repartParam = DParam ( PreGroupByReserialization = 0, NumPartitions = finalPartitions )
         let dkvRepart = dkvGen.RepartitionP(repartParam, partitionFunc)
         let t3 = DateTime.UtcNow
         //let verifyCount = dkvGen |> DSet.fold (fun cnt x -> cnt + 1) (fun a b -> a + b) 0
+        let foldFunc = (fun cnt x ->
+            cnt + 1
+        )
```
BrunoBozza: It seems like foldFunc is dead. If you'd like to keep it around, you could comment it out and add a comment explaining why it could be useful.

src/CoreLib/DSeq.fs
sanjeevm0, line 901: Yes, flow control is disabled here, as network handles all flow control

src/tools/tools/configuration.fs
sanjeevm0: Changed both to Stream

samples/SortBenchmark/DiskHelper.fs
```diff
@@ -41,13 +41,12 @@ open Prajna.Tools
...
 open Prajna.Core
-//open Prajna.Service.CoreServices
 
 [<Serializable>]
 type StorageProfile (profileName : string) =
     //member val diskProfile = [|"c:\\sortbenchmark\\";"c:\\sortbenchmark\\";"c:\\sortbenchmark\\";"c:\\sortbenchmark\\";"d:\\sortbenchmark\\";"e:\\sortbenchmark\\";"f:\\sortbenchmark\\"|] with get
-    member val drives = DriveInfo.GetDrives()
-    member val diskProfile = DriveInfo.GetDrives() |> Array.filter(fun di -> di.DriveType = DriveType.Fixed)  |> Array.filter(fun di -> di.TotalSize > 100000000000L) |> Array.map(fun di -> Path.Combine(di.Name,"sortbenchmark").ToLower()+"\\")
+    //member val drives = DriveInfo.GetDrives()
+    member val diskProfile = null with get,set
```
BrunoBozza, line 49: Public properties should be use PascalCasing.
Also, what's the type of diskProfile? I was thinking StorageProfile, but that doesn't allow the null literal. Maybe add a type annotation?
Does it need to be initialized with null? In internal F# code, I would make this an option, but since it's a public type, maybe we could use a "NullProfile" object? This would avoid null checking everywhere else.

samples/SortBenchmark/DiskHelper.fs
```diff
@@ -291,3 +299,85 @@ type DiskHelper(records:int64) =
                 x.sorteddataSP.Save() 
                 b <- true
                 x.sorteddataSP.lock := 0
+
+
+
+[<AllowNullLiteral>]
```
BrunoBozza, line 305: General comment: I am against "null" whenever possible. In F#, there's option. For public interface code, if really there's no way to construct objects that are "complete", we can use the null object pattern: https://en.wikipedia.org/wiki/Null_Object_pattern. Unfortunately, C# doesn't have a "[DisallowNullLiteral]" attribute, so ultimately one is always exposed. But generally avoiding null pointers at runtime will go a long way towards avoiding NullReferenceException. That's one of the most common kinds of exceptions, and it's a whole class of bugs that can largely mitigated.

samples/SortBenchmark/DiskHelper.fs
```diff
@@ -291,3 +299,85 @@ type DiskHelper(records:int64) =
...
+type RollingFileMgr(records:int64) =
```
BrunoBozza, line 306: Public properties and fields should be PascalCased.

samples/SortBenchmark/DiskHelper.fs
```diff
@@ -291,3 +299,85 @@ type DiskHelper(records:int64) =
...
+        new FileStream(fp,FileMode.Create)
+
+
+    member x.EnqueueFileHandle(fh:FileStream, size:int) =
+        x.fileHandleQ.Enqueue(fh,size)
+
+
+
+[<AllowNullLiteral>]
+type SingleThreadExec1() =
```
BrunoBozza, line 347: There's almost never a reason for a type with an empty constructor to have an [<AllowNullLiteral>]. To delay initialization, we can always just use a Lazy<'T>, or just create the object if initialization is not an issue.

samples/SortBenchmark/SortBenchmark.fs
```diff
@@ -43,40 +43,23 @@ open Prajna.Tools
...
 let Usage = "
     Usage: Benchmark performance for distributed sort. \n\
     Command line arguments:\n\
     -in         Copy into Prajna \n\
-    -out        Copy outof Prajna \n\
     -dir        Directory where the sort gen file stays \n\
-    -num        Number of remote instances running \n\
-    -sort       Executing Sort (1: 1-pass sort, 2-2 pass sort) \n\
-    -nump N     Number of partitions \n\
+    -nump       Number of partitions in the second stage repartition\n\
```
BrunoBozza, line 59: We should start using a proper command line parsing library. Ken was working in VMHub with the same one we used in TLC/AzureML Algos. It's pretty good. I'm sure there are others. This should not block this pull, of course, as there's tons of code that's like this and we don't need to change it now. But going forward, we should move towards something like that.

samples/SortBenchmark/SortBenchmark.fs
```diff
@@ -97,12 +80,19 @@ module Interop =
...
         let pinedbuffer = bufferHandle.AddrOfPinnedObject()        
         stdqsort(pinedbuffer, len / dim, dim)
         bufferHandle.Free()
 
+    let STLqsort (buffer:byte[], dim:int) = 
+        let bufferHandle = GCHandle.Alloc(buffer,GCHandleType.Pinned);
+        let pinedbuffer = bufferHandle.AddrOfPinnedObject()        
+        stdqsort(pinedbuffer, buffer.Length / dim, dim)
+        bufferHandle.Free()
+
```
BrunoBozza, line 95: Use try/finally to free the handle.

samples/SortBenchmark/SortBenchmark.fs
```diff
@@ -141,148 +131,34 @@ module Interop =
...
-                                        fileHandle.Close()
-                            )
-        ()
-
-    member internal x.CheckAndWriteToFile() = 
-
-        let fileReadyToWrite = dataBufSize.ToArray() 
-                                |> Seq.filter (fun kv -> !(kv.Value) > x.WriteBlockSize  ) 
-                                |> Seq.map (fun kv -> kv.Key)
+type RepartitionStage = StageOne=1 | StageTwo=2
```
BrunoBozza, line 135: Does this need to be public?

samples/SortBenchmark/SortBenchmark.fs
```diff
@@ -390,98 +266,310 @@ type RemoteFunc( filePartNum:int, records:int64, _dim:int , partNumS1:int, partN
...
-                        }
-        r
 
-    member x.ReadFilesfake parti= 
+
+    member x.GetReadFileBuf() = 
+            let byt = ref Unchecked.defaultof<_>
+            let bnewbuf = ref false
+            if not (RemoteFunc.sharedMem.TryDequeue(byt)) then
+                if (Interlocked.Increment(RemoteFunc.sharedMemSize) < 300) then
```
BrunoBozza, line 278: General comment: There's no need for the extra indirection with "ref". So long as you don't capture the variables (byt and bnewbuf) in closures, you can use just "let mutable" and "&byt" when passing to TryDeque. Probably doesn't make much difference in most code, but just fyi.

samples/SortBenchmark/SortBenchmark.fs
```diff
@@ -390,98 +266,310 @@ type RemoteFunc( filePartNum:int, records:int64, _dim:int , partNumS1:int, partN
...
+                        let t1 = DateTime.UtcNow
+                        let bHasBuf = ref true
+                        let sr = new StreamReader<byte>(buffer,0L)
+
+                        let bRemainRecord = ref false
+                        let remainBuf = ref Unchecked.defaultof<byte[]>
+                        let remainpos = ref 0
+                        let remainLen = ref 0
+
+                        while !bHasBuf do
```
BrunoBozza, line 468: Seems you really care about perf. I think you'll see some benefit from "let mutable", not to mention less noisy syntax.

samples/SortBenchmark/SortBenchmark.fs
```diff
@@ -390,98 +266,310 @@ type RemoteFunc( filePartNum:int, records:int64, _dim:int , partNumS1:int, partN
...
+                            let (buf, pos,len) = sr.GetMoreBuffer()
+                            if len > 0 then
+                                let idx = ref pos
+
+                                if !bRemainRecord then
+                                    if (!remainLen - !remainpos >=2) then
+                                        let index = (((int) (!remainBuf).[!remainpos]) <<< 8) + ((int) (!remainBuf).[!remainpos + 1])
```
BrunoBozza, line 475: No need to parenthesize the "int" in a typecast, since it's a regular function.

samples/SortBenchmark/SortBenchmark.fs
```diff
@@ -552,36 +631,103 @@ type RemoteFunc( filePartNum:int, records:int64, _dim:int , partNumS1:int, partN
...
+
+    // repartition, use per-allocated memory. For testing and comparing to MemoryStreamB only, cannot be used in remote server
+    member x.RepartitionSharedMemory (stage:RepartitionStage) (buffer:byte[], size:int) = 
         if size > 0 then
             
             let retseq = seq {
                         let partionBoundary =
-                            if stage = 1 then
+                            if stage = RepartitionStage.StageOne then
                                 stageOnePartionBoundary
```
BrunoBozza, line 720: If the stage is never sent over the wire, there's an advantage to _not_ making it an enum (by ommiting the "= 1", "= 2" in the stage definition). The advantage is that the "match" keyword will check that the alternatives are exhaustive at compile time. As this is, if a "StageThree" is ever introduced, we'll have to go hunting for where to update. If we use an F# union and "match", the compiler does that for us.

src/CoreLib/DSeq.fs
```diff
@@ -1011,10 +1041,14 @@ and [<AllowNullLiteral>]
...
                 // Add serial and numElems as prefix. 
-                let msPrefix = new MemStream( 128 )
+                //let msPrefix = new MemStream( 128 )
+                let msPrefix = ms.GetNew()
                 msPrefix.WriteInt64( meta.Serial )
                 msPrefix.WriteVInt32( meta.NumElems ) 
-                let msCombine = ms.InsertBefore( msPrefix )
+                //let msCombine = ms.InsertBefore( msPrefix )
+                ms.InsertBefore(msPrefix) |> ignore
+                ms.DecRef()
```
BrunoBozza, line 1050: Can we wrap the GetNew()/DecRef() pair into a "using" type function, so the DecRef() is guaranteed. Something like:

`````` F#
let withNew (f: stream -> 'a) ms = 
    let x = ms.GetNew() 
    let ret = f x
    ms.DecRef()
    ret```
``````

File: samples/SortBenchmark/DiskHelper.fs (hunk 11 above)
Comment (hongzhili) [inline samples/SortBenchmark/DiskHelper.fs:49]: diskProfile is an array. 
It has to be initialized at remote container in order to get hard drive configuration of the remote machine (not client machine). And yes, a type annotation or comment is needed here.

File: src/CoreLib/network.fs [hunk 23]
```diff
@@ -170,55 +170,59 @@ type NetworkPerformance() =
...
+                0L
         else
-            x.count
+            if (x.buildHeader) then
+                8L + x.ms.Length - x.startPos
+            else
+                x.ms.Length - x.startPos
+
+    /// release the underlying memstream
+    member x.Release() =
```
BrunoBozza, line 127: Why not standard IDisposeable with a call to SupressFinalizer in the Dispose() implementation?

src/CoreLib/network.fs
```diff
@@ -1001,29 +1099,183 @@ type [<AllowNullLiteral>] NetworkCommandQueue() as x =
...
+        ms.WriteByte(byte(command.Verb))
+        ms.WriteByte(byte(command.Noun))
+        Logger.LogF( LogLevel.ExtremeVerbose, (fun _ -> sprintf "Send command to %s len: %d ack: %d" x.EPInfo cmdLen x.RcvdCommandSerial))
+        ms.WriteUInt16(Operators.uint16(x.RcvdCommandSerial))
 
     member private x.UpdateHeader(cmd : NetworkCommand) =
-        let rcvdSerialBuf = BitConverter.GetBytes(Operators.uint16(x.RcvdCommandSerial))
-        Buffer.BlockCopy(rcvdSerialBuf, 0, cmd.buffer, 6, 2)
+        let pos = cmd.ms.Position
+        cmd.ms.Seek(6L, SeekOrigin.Begin) |> ignore
```
BrunoBozza: Why 6? Maybe use symbolic constants?

samples/SortBenchmark/DiskHelper.fs
hongzhili, line 305: I agree. Actually, this type may be removed later since we changed the implementation.

src/CoreLib/DSeq.fs
BrunoBozza, line 1050: Same for AddRef.

BrunoBozza:
Phew! I'm done. I'll update my generic serialization to use this new code once it goes in. In general, there should be no reason for anyone to write a "custom" encoder/decoder for arrays/matrices unless they know something special about their data (e.g.: image, sound, etc). We could enable general gzip/lzw encoding/decoding as an option, which would go a long way for text. If we also add something lossless for System.Drawing.Bitmap and whatever the .Net PCM waveform object is, we should be in good shape for many workloads with quite good performance.

src/CoreLib/DSetGenerics.fs
sanjeevm0: Done

src/CoreLib/deploy.fs
sanjeevm0, line 222: I think it is fine to use this as memory is controlled by the shared stack.  Using higher number may in some cases increase delay on per message basis, but most of our current workload is not sensitive to this.

src/CoreLib/deploy.fs
sanjeevm0, line 261: 4GB is fine as there is additional control of using certain percentage of available RAM.

samples/DistributedSort/DistributedSort.fs
sanjeevm0: done

src/CoreLib/DSeq.fs
sanjeevm0, line 1050: Yes, Jin is also taking a look at this.  This will eventually happen.

src/CoreLib/network.fs
sanjeevm0, line 127: I have added a IDisposable, however I have left Release in also.  It is a bit hard to wrap NetworkCommand in a using statement as it is in a queue, and when processing is done, it needs to be released.

samples/SortBenchmark/DiskHelper.fs
BrunoBozza, line 49: Can it be initialized with a Lazy<'a[]>? If this is meant to be accessed only at the remote machine, a Lazy has another advantage: it let's you throw an informative exception in case one tries to access it at the client. Could be as simple as 'failwith "Shouldn't be accessing this at the client"', which is way better than a NullReferenceException, and making everyone check for null everywhere.
To be clear: I only propose implementing this before the pull if it is trivial and doesn't add any delay to the commit. Otherwise, we should pull this as is and change (or not) later. My general take on code reviews is that, as long as the code is useful and not harmful outside its context (and this clearly qualifies), commited code is always better than code sitting in your repo.

hongzhili, line 49: Thanks Bruno for the suggestions. I agree. Since SortBenchmark is an individual sample project. Let's just merge this pull request if all other issues are addressed and I can integrate the comments after it is merged. Actually, it's not easy for me to make any changes right now since it's in Sanjeev's branch.

PR #44, opened 2015-09-15
Supporting memory-copy-based serialization of matrices

BrunoBozza:
This adds support for generic, memory-copy-based serialization of matrices and higher rank arrays. Other than the update to zero-object replication by Jin on tasks.fs, it's all contained to generic serialization code.

### DIFF
--- a/src/CoreLib/task.fs
+++ b/src/CoreLib/task.fs
@@ -1459,7 +1459,9 @@ and [<AllowNullLiteral; Serializable>]
                                         null
                                     else 
                                         // Start pos will not be the end of stream, garanteed by state not null 
-                                        new MemStream( buf, pos, length, false, true )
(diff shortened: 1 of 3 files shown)


zhuweirong:
Squash the two commits into one?

BrunoBozza:
Squashed! Had to tell git what to rebase: "git rebase HEAD~1" worked.

zhuweirong:
It seems there are 5 commits now  ..... the rebase is also rewriting history of some other existing commits.  In this case, it might be easier to just do a  "git reset upstream/master"  (assume "upstream" refers to MSRCCS/master), then do a force push. It will appear as a single commit.

BrunoBozza:
Ok. Single commit now. Hadn't seen the mess that the rebase made.

PR #48, opened 2015-09-16
Adding matrix vector multiply sample

BrunoBozza:
Adding matrix vector multiply sample


zhuweirong:
Cannot merge, GitHub says "This branch is out-of-date with the base branch"

BrunoBozza:
Someone updated the docs in the meantime (Prajna_v1.docm and Prajna_v1.pdf). I just pulled and pushed again. Try now.

zhuweirong:
Hmm, it seems due to I changed the setting last night to "protect" the master branch. As a result, it has stronger requirement -- e.g. your branch has to be up to date, even there when there is no conflict. I have temporarily disabled "protect" feature.  We need to think whether we want it. It can protect us from "push --force" or accidentally delete the branch etc. Seems to be useful, but make the pull request more difficult. Any opinion?

BrunoBozza:
I think keeping protect is fine. It will only be a problem in cases like now, when we are pushing roughly at the same time. When something happens, we can just pull and push again like I just did.

zhuweirong:
It seems that if I enable "protect" but not enable "status check", it will allow merge even the source branch is not up-to-date as long as there's no conflicts.

PR #50, opened 2015-09-17
Deleting useless HDD store from Matrix sample

BrunoBozza:
Was trying to hide DSet load latency, but it doesn't seem to make a difference, so simplifying sample.


zhuweirong:
Take a look at WordMatch, it uses "store" then "loadSource".  When the matrix becomes large, this may make difference compared to using distribute.  "Distribute" as it is now, capture the whole matrix as part of the closure, and thus send the whole matrix to all nodes, then each node takes part of it.  "Store" on the other hand, only send the data that is need for each node.  The downside of "store" is that it persists the data into disk.  Ideally, "Distribute" should be reimplement-ed, but it's non-trivial.

PR #52, opened 2015-09-18
Deployment Scripts for domain cluster

zhuweirong:
1. Deployment scripts for windows domain clusters
2. Support plain text list file - lst (so user does not have to use inf file)
3. Support password with the lst file
4. New algorithm on system saved cluster metadata
5. Fix a few issues/bugs

There're still a few TODO's
- Add a readme file with the scripts
- Include the scripts to the nuget package
- When the cluster requires passwd but user does not provide it, the exception message is not friendly.

### DIFF
--- a/src/CoreLib/client.fs
+++ b/src/CoreLib/client.fs
@@ -46,32 +46,12 @@ type internal ClientLauncher() =
     inherit System.MarshalByRefObject() 
     static let Usage = "PrajnaClient \n\
         Command line arguments:\n\
-        -allowad Allow AppDoamin execution \n\
-        -master  MASTER_INFORMATION_FILE : master information file\n\
-        -dir     WORKING_DIRECTORY :  working directory\n\
-        -mem     MAX_MEMORY_LIMIT : Maximum Memory Limit in MB\n\
-        -totaljob   TOTAL_JOB_LIMIT : set number of jobs that can be executed by Prajna in parallel\n\
-        -local   start on local machine\n\
-        -usealldrives use all drives for data (including system drive)\n\
+        -dirlog  the dir for the log files\n\
+        -pwd     the passwd used for job request authentication\n\
+        -mem     maximum memory size in MB allowed for the client\n\
+        -port    the port that is used to listen the request\n\
+        -jobport the range of ports for jobs\n\
         "
-
-    static let TryRemoteAccess() = 
-        let curAcct = System.Security.Principal.WindowsIdentity.GetCurrent()
-        Logger.Log( LogLevel.Info, ( sprintf "Current user: %s (IsAuthenticated: %A)" (curAcct.Name) (curAcct.IsAuthenticated) ))
-        //    for claim in curAcct.Claims do 
-        //        Logger.Log(LogLevel.Info, ( sprintf "Type %A Subject:%A Issuer:%A value:%A " claim.Type claim.Subject claim.Issuer claim.Value ))
-        try
-            let fname = """\\OneNet20\c$\Prajna\Log\JinL\test_remotewrite.txt""" 
-            let fStream = new FileStream( fname, FileMode.Create, Security.AccessControl.FileSystemRights.FullControl, 
-                                        FileShare.ReadWrite, (1<<<20), FileOptions.Asynchronous )
-            let enc = System.Text.UTF8Encoding().GetBytes( "To test if remote write can succeed, cow jump over the moon" )
-            fStream.Write( enc, 0, enc.Length )
-            fStream.Flush()
-            fStream.Close()
-            Logger.Log( LogLevel.Info, ( sprintf "Remote writing succeeds" ))
-        with
-        | e -> 
-            Logger.Log( LogLevel.Info, ( sprintf "Remote writing fails with exception %A" e))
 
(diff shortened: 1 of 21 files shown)


src/CoreLib/deploy.fs
```diff
@@ -139,7 +139,7 @@ type DeploymentSettings() =
             drives |> Seq.find( fun drive -> not (DeploymentSettings.ExcludedDrives.ContainsKey( drive )) )
     static member val internal GetFirstDataDrive = DeploymentSettings.GetFirstDataDriveImpl with get, set
     /// Directory at remote node that holds Prajna operating state. 
-    static member val LocalFolder = Path.Combine(DeploymentSettings.RootFolder, "OneNet") with get
+    static member val LocalFolder = Path.Combine(DeploymentSettings.RootFolder, "Prajna") with get
```
zhuweirong, line 142: For test clusters with existing data, on each machine, I will create a symbolic link Prajna that points to original folder

src/CoreLib/deploy.fs
```diff
@@ -159,7 +159,7 @@ type DeploymentSettings() =
     static member val internal MasterPort = 1080 with get, set
     static member val internal ControlPort = 1081 with get, set
     /// Directory at remote node that holds Prajna data files
-    static member val DataFolder = @"OneNetData" with get
+    static member val DataFolder = @"PrajnaData" with get
```
zhuweirong, line 162: For test clusters with existing data, on each machine, I will create a symbolic link PrajnaData that points to original folder

src/CoreLib/client.fs
```diff
@@ -225,13 +206,20 @@ type internal ClientLauncher() =
                 null
         let listener = Listener.StartListener()
         let monTimer = ThreadPoolTimer.TimerWait ( fun _ -> sprintf "Task Queue Monitor Timer" ) listener.TaskQueue.MonitorTasks DeploymentSettings.MonitorIntervalTaskQueueInMs DeploymentSettings.MonitorIntervalTaskQueueInMs
+        
         // set authentication parameters for networking
+        Logger.Log( LogLevel.Info, (sprintf "Authentication parameters: pwd=%s keyfile=%s keyfilepwd=%s" pwd keyfile keyfilepwd ) )
```
sanjeevm0: Do we want to log passwords directly in logfile (they will be visible in the clear)?

src/CoreLib/network.fs
```diff
@@ -287,9 +287,9 @@ type [<AllowNullLiteral>] NetworkCommandQueue() as x =
             if (isIp) then
                 NetworkCommandQueue.EndResolveDNS(ipAddr, true, (x, port))
             else
-                let success = LocalDNS.BeginDNSResolve(x.MachineName, NetworkCommandQueue.EndResolveDNS, (x, port))
-                if (not success) then
-                    x.MarkFail()
+            let success = LocalDNS.BeginDNSResolve(x.MachineName, NetworkCommandQueue.EndResolveDNS, (x, port))
```
sanjeevm0: Indentation seems incorrect after else statement?

src/CoreLib/network.fs
```diff
@@ -307,6 +307,7 @@ type [<AllowNullLiteral>] NetworkCommandQueue() as x =
         x.RequireAuth <- requireAuth
         x.MyGuid <- myguid
         if (requireAuth) then
+            Logger.LogF(LogLevel.MildVerbose, fun _ -> sprintf "LoopbackConnect Auth: pwd=%s" pwd)
```
sanjeevm0: Same comment regarding logging of passwords in clear.

src/CoreLib/network.fs
```diff
@@ -1340,6 +1341,12 @@ and [<AllowNullLiteral>] NetworkConnections() as x =
         if Utils.IsNotNull dirpath then
             FileTools.DirectoryInfoCreateIfNotExists (dirpath) |> ignore
 
+    /// Initialize authentication using a shared secret password 
+    member x.InitializeAuthentication (pwd : string) : unit =
+        let keyFile =  Path.Combine( [| DeploymentSettings.LocalFolder; "Keys";  Guid.NewGuid().ToString("D") |] )
+        let keyFilePwd = Guid.NewGuid().ToString("D")
+        x.InitializeAuthentication(pwd, keyFile, String.Empty)
```
sanjeevm0: keyfilePwd cannot be generated, has to be passed in.  Plus it seems you are passing String.Empty, so perhaps it is best to just take out the line "let keyFilePwd = Guid.NewGuid().ToString("D") "

src/CoreLib/client.fs
zhuweirong: No, we do not and should not. It was debug code, I will clean it.

src/CoreLib/network.fs
zhuweirong: will fix

src/CoreLib/network.fs
zhuweirong: agree

src/CoreLib/network.fs
zhuweirong: I found that if I leave it to empty, it does not work somehow. That's why I put something to work it around for now. I will double check.

PR #53, opened 2015-09-18
Replacing functions & code reorganization:

sanjeevm0:
1. Remove qsort, replace with qsort_s …  
2. Remove memcpy, replace with memcpy_s
3. Remove RijndaelManaged, replace with AesCryptoServiceProvider with 256 bit key
4. Reorganize custom serialization code so it is in one place
5. Fix bug in tests/tools/tools/mem.fs - testConstructor unit test


samples/SortBenchmark/qsort/qsort.cpp
```diff
@@ -35,10 +34,10 @@ bool compareb(const void * a, const void * b)
 }
 
 extern "C" __declspec(dllexport)
-void stdqsort(void* buf, int len, int dim)
+void stdquicksort(void* buf, int len, int dim)
 {
 	dim = d;
-	qsort(buf, len, dim, compare);
+	qsort_s(buf, len, dim, compare, null);
```
zhuweirong, line 40: null -> nullptr

sanjeevm0, line 40: AFAIK, nullptr is only C++?

zhuweirong, line 40: "extern "C"" makes a function name in C++ have 'C' linkage (thus no name mangling).  But this code is still C++, right?

sanjeevm0, line 40: Probably.  It is not actually being actively compiled, so I am not verifying correctness at this point.  

Perhaps we can just stash the code in a branch at this point?

Date: Wed, 23 Sep 2015 14:36:02 -0700

PR #55, opened 2015-09-18
Add Json serialization support in StreamExtension and StringExtension

leizhangcn:


zhuweirong:
1. Should have squashed the two commits into one. 
2. This change add a new framework dependency (System.Runtime.Serialization") for Prajna.Tools.  It means we need to update the nuget package spec too. 

Next time, please give us some time to review before merging the pull request?

leizhangcn:
Got it. Will pay attention next time and let you or Jin to review before merge.

Any guideline when to squash two commits and when to keep both? I am still a bit confused about this.

Thanks,
Lei

PR #57, opened 2015-09-18
By JinL, use Stream instead of StreamBase in Prajna.Service

leizhangcn:
By JinL, use Stream instead of StreamBase in Prajna.Service


zhuweirong:
Please squash the two commits into a one? Otherwise, the "merge" commit would appear in commit history too.

leizhangcn:
Sure, will pay attention to this guideline next time.

PR #61, opened 2015-09-23
Unregister callbacks and unallocate blobs when job ends

zhuweirong:


zhuweirong:
Jin has addressed this issue in the feature of cancellation, which will be done soon.

PR #62, opened 2015-09-25
Adding Fletcher64 hashing to replace SHA256

BrunoBozza:
SHA256 was way overkill what we are doing. This Fletcher64 implementation is more than sufficient for checksuming, over 50x faster, and does around 5GB/s in my machine.

### DIFF
--- a/src/CoreLib/dependency.fs
+++ b/src/CoreLib/dependency.fs
@@ -55,7 +55,9 @@ type internal JobDependency() =
     /// we use the last 8B of Hash as signature
     member x.ComputeHash( buf: byte[] ) = 
         if (Utils.IsNull x.Hash) then
-            x.Hash <- HashLengthPlusByteArray( buf )
+            x.Hash <- 
+                use hasher = new System.Security.Cryptography.SHA256Managed()
+                HashLengthPlusByteArray( hasher, buf )
         x.Hash
 
     /// Compute hash based on current file 
@@ -257,7 +259,6 @@ type JobDependencies() =
             ms.WriteBytes( dep.Hash ) |> ignore 
         let (buf, pos, cnt) = ms.GetBufferPosLength()
         x.Hash <- ms.ComputeSHA256(int64 pos, int64 cnt)
-        //x.Hash <- HashByteArrayWithLength( ms.GetBufferPosLength() )
         ms.DecRef()
         x.Hash
 
--- a/src/tools/tools/hash.fs
+++ b/src/tools/tools/hash.fs
@@ -1,0 +1,197 @@
+﻿namespace Prajna.Tools
+
+#nowarn "9"
+    
+open System
+open System.Runtime.InteropServices
+open System.Security.Cryptography
+
+open FSharp.NativeInterop
+
+module Hash =
+
+    open System.Collections.Generic
+
+//    let inline safeItersWithoutOverflow maxShortAsLong = 
+//        let sums = 
+//            seq {
+//                let mutable sum1 = maxShortAsLong
+//                let mutable sum2 = maxShortAsLong
+//                while true do
+//                    sum1 <- sum1 + maxShortAsLong
+//                    sum2 <- sum2 + sum1
+//                    sum1 <- sum1 + maxShortAsLong
+//                    sum2 <- sum2 + sum1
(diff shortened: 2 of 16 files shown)


zhuweirong:
Does not pass tests. Many tests fail with messages like "Job dependency hash 2E31FFB457312015 doesn't match the hash generated by Prajna 4CA8DC357DF020914E1FFF3D89400848B1F330A59608C0A5E8EF134C17237610"

src/tools/tools/hash.fs
```diff
@@ -0,0 +1,173 @@
...
+
+#nowarn "9"
+    
+open System
+open System.Runtime.InteropServices
+open System.Security.Cryptography
+
+open FSharp.NativeInterop
+
+module Hash =
```
zhuweirong, line 11: "internal",  I don't assume we want to expose it to Prajna users as public APIs.

src/tools/tools/hash.fs
```diff
@@ -0,0 +1,221 @@
...
+//            member this.Finish(): byte [] = 
+//                inner.TransformFinalBlock([||], 0, 0) |> ignore
+//                inner.Hash
+//            member this.ComputeHash(buffer: byte[], start: int, len: int): byte[] =
+//                inner.ComputeHash(buffer, start, len)
+//            member this.Dispose() = inner.Dispose()
+//
+////        interface System.Security
+
+module Hash =
```
zhuweirong: internal?

src/CoreLib/assembly.fs
```diff
@@ -227,7 +227,7 @@ and [<AllowNullLiteral>]
             x.Pack( ms )  
             // x.Hash <- HashByteArrayWithLength( ms.GetBufferPosLength()  )
             let (ms, pos, len) = ms.GetBufferPosLength()
-            x.Hash <- ms.ComputeSHA256(int64 pos, int64 len)
+            x.Hash <- ms.ComputeDefaultHash(int64 pos, int64 len)
```
jinlmsft: I have discussed with Bruno. Let's use strong hash (SHA256) for assembly and files. Otherwise, we are open ourselves to attack scenarios. Specifically, let's use strong hash for BlobKind.AssemblyManagedDLL, BlobKind.AssemblyUnmanagedDir, BlobKind.JobDependencyFile

src/CoreLib/dependency.fs
```diff
@@ -250,20 +250,20 @@ type JobDependencies() =
     member val EnvVars : List<string*string> = new List<string*string>() with get
     
     /// Job File Dependencies Hash
-    member internal x.JobDependencySHA256() =
+    member internal x.JobDependencyDefaultHash() =
```
jinlmsft: Use Strong Hash.

BrunoBozza: Yep. I also have a few renamings to make. "Default" is not a very good name. But what else? I think I'll call it "Checksum" instead of Hash, to make it clear what it is, and not be bound to the Fletcher64 algorithm.

zhuweirong:
I assume that you have verified that the authentication described here (https://github.com/MSRCCS/Prajna/wiki/Deploy-Prajna) still works fine?

BrunoBozza:
No, I did not check that. When was it last tested? I just want to make sure, before I go test it, that any difficulties are due to my changes, not the massive changes that Jin just added in the last week or so.

BrunoBozza:
What's up with the AppVeyor build? It failed with the following:

Running build failed.
Error:
System.ComponentModel.Win32Exception (0x80004005): Access is denied
   at System.Diagnostics.Process.Kill()
   at FSI_0001.Build.f@1(Process p) in C:\projects\prajna-rbpft\build.fsx:line 424
   at FSI_0001.Build.runTests(String target) in C:\projects\prajna-rbpft\build.fsx:line 423
   at FSI_0001.Build.clo@427-29.Invoke(Unit _arg2) in C:\projects\prajna-rbpft\build.fsx:line 427
   at Fake.TargetHelper.runSingleTarget(TargetTemplate`1 target) in C:\code\fake\src\app\FakeLib\TargetHelper.fs:line 

Seems unrelated to my change.

zhuweirong:
I'm looking at it ....

PR #64, opened 2015-09-26
Implement JobLifeCycle Object, use IDispoble for Task, Job, Stream.

jinlmsft:
This is a big aggregated change on the Prajna Project. This pull request is not fully ready to check-in yet. But I will use this pull request to initiate a code-review request, and
disucss the major changes incorporated in this version.

The main functions implemented are with regard to distributed exception handling and distributed resource management. Our goal will be: 1) Job may throw exception at remote, and all exception
will be serialized and shown at local. 2) We should properly implement handling of IDisposable object throughout our project.

1) JobLifeCycle object.

Each Prajna action (read, fold, etc.) is now associated with a GUID, which is further associated with a JobLifeCycle object. The JobLifeCycle governs the life cycle of a Prajna job,
and is responsible to release any resource used by Job (at App, Daemon, or at Container).

You can register callback (event) on the JobLifeCycle as follows.

OnDispose/OnDisposeFS: function to be called when the job is to be disposed. (only once right before job is disposed).
OnException/OnExceptionFS: function to be called when an exception is encountered. (may be called multiple times if multiple exception is encountered)
OnCancellation/OnCancellationFS: function to be called when the job is cancelled. (only once per job)

JobLifeCycle implements IDisposable interface.

2) SingleJobActionApp/SingleJobActionContainer/SingleJobActionDaemon

JobLifeCycle should not usually be used. Instead, function that plan to do something on the job should attempt to grab a SingleJobActionApp/SingleJobActionContainer/SingleJobActionDaemon object.
SingleJobAction\* object also implements IDisposable interface. If a SingleJobAction\* object has been successfully grabbed, the subsequent operation can safely use the resource of the job
without fearing that the resource is disposed. The operating logic is that when a SingleJobAction\* object will try to increment the reference count of JobLifeCycle when it is created, and
check whether the job is still properly running. If not, try to grab SingleJobAction\* object will return a null pointer, and the subsequent operations should not be executed.
When it is disposed, it will attempt to decrement the reference count. When Job is cancelled and the JobLifeCycle object is disposed, it will also decrement the reference count, and mark
the Job as being disposed. Only when the reference count goes to negative, the process will kick in to dispose the job resources.

3) ThrowExceptionAtCallback/EncounterExceptionAtCallback
   /ThrowExceptionAtContainer/EncounterExceptionAtContainer:

These functions are being used to catch exception at remote, aggregate exception, and throw at the local. Ultimately, we should move to a stage that most exceptions at remote (including our
own programming error) should be manifested at local.

4) Job/Task: disposable.

Both Job and Task become disposable object. Thus, they can properly free their resource if they fail.

5) MemStream/StreamBase<byte>, etc..

I start to use the disposable feature of MemStream/StreamBase<byte> to govern stream memory allocation and resource release. My thought is as follows.
a) Application that allocates the stream should be responsible to free its resource. That is, we should be able to
    use ms = new MemStream() ... for most of our jobs.
b) If other function needs to retain a copy of the stream, it should use .Replicate() to get a duplicated copy of the stream.
c) We should try to get rid of .AddRef() and .DecRef() management of the memory stream.


src/tools/tools/LoggerCSharp.fs
```diff
@@ -47,6 +47,18 @@ type Logger private() =
         if Logger.LoggerProvider.IsEnabled(logId, logLevel) then
             Logger.LoggerProvider.Log((logId, logLevel, messageFunc.Invoke()))
 
+    /// Log message generated by "messageFunc" using "JobId", 
+    /// "messageFunc" is evaluated only if "logLevel" is less than or equal to the log level for "logId"
+    static member inline LogF(jobID : Guid, logLevel : LogLevel,messageFunc : Func<string>) =
```
zhuweirong, line 52: This is to be used by C#, probably no need to "inline" to be consistent with other APIs.

src/tools/tools/Logger.fs
```diff
@@ -297,6 +311,16 @@ type internal DefaultLogger () =
         member this.Log ((logId : string, logLevel : LogLevel, message : string)) =
             // DefaultLogger does not support "log id" yet
             EmitLogEntry(logLevel, message)
+        
+        member this.Log ((logId : string, jobID: Guid, logLevel : LogLevel, message : string)) =
```
zhuweirong, line 328: The "Logger" class below should have a non-functional flavor "Log" function that takes JobID as parameter

src/tools/tools/StringTools.fs
```diff
@@ -60,6 +61,16 @@ type StringTComparer<'T when 'T:equality>=
             else
                 (17 * 23 + this._comp.GetHashCode( xstring )) * 23 + xval.GetHashCode()
 
+/// <summary>
+/// Construct a comparer that uses Object.ReferenceEquals to compare object. 
+/// </summary>
+type ReferenceComparer<'T>()=
```
zhuweirong: why it's in StringTools.fs? seems not specific to string.  Maybe in utils.fs?  Also, should it be internal?

src/tools/tools/extension.fs
```diff
@@ -356,4 +356,19 @@ type StringExtension =
...
     static member DeserializeFromJson( s: String ): 'T = 
         use ms = new MemoryStream(System.Text.Encoding.UTF8.GetBytes(s))
-        ms.ReadJson()
\ No newline at end of file
+        ms.ReadJson()
+    /// Helper function to pack a remote exception 
+    /// See: http://blogs.msdn.com/b/brada/archive/2005/03/27/402801.aspx
+    [<Extension>]
+    static member WriteException( x: Stream, ex: Exception ) = 
+        let fmt = System.Runtime.Serialization.Formatters.Binary.BinaryFormatter()
```
zhuweirong, line 364: Should it use the formatter that Bruno wrote?

src/tools/tools/extension.fs
```diff
@@ -356,4 +356,19 @@ type StringExtension =
...
+        fmt.Serialize( x, ex )
+    /// Helper function to pack a remote exception 
+    /// See: http://blogs.msdn.com/b/brada/archive/2005/03/27/402801.aspx
+    [<Extension>]
+    static member ReadException( x: Stream ) = 
+        let fmt = System.Runtime.Serialization.Formatters.Binary.BinaryFormatter()
```
zhuweirong, line 370: Use the new formatter Bruno provided?

src/CoreLib/DSeq.fs
```diff
@@ -89,23 +90,26 @@ and internal DStreamForwardDependencyType =
     /// Multicast a blob to all destinations in the Mapping
     | MulticastToNetwork of DependentDStream
 and internal DStreamFactory() = 
-    inherit HashCacheFactory<DStream>()
-    static let collectionByName = ConcurrentDictionary<_,_>(StringTComparer<int64>(StringComparer.Ordinal)) 
-    // Retrieve a DStream
-    static member ResolveDStream( hash ) = 
-        DStreamFactory.Resolve( hash ) 
+    static let collectionByJob = ConcurrentDictionary<Guid,(int64 ref)*ConcurrentDictionary<_,_>>()
```
zhuweirong, line 96: add some comments to explain about the value type  (int64 ref) \* ConcurrentDictionary<_,_>, what are they?

src/CoreLib/DSeq.fs
```diff
@@ -532,71 +537,88 @@ and [<AllowNullLiteral>]
     /// Return: 
     ///     false: blocking, need to resend the command at a later time
     ///     true: command processed. 
-    member internal x.ProcessJobCommand( queuePeer:NetworkCommandQueue, cmd:ControllerCommand, ms:StreamBase<byte> ) = 
-        let peeri = x.Cluster.SearchForEndPoint( queuePeer )
-        Logger.LogF( LogLevel.WildVerbose, ( fun _ -> sprintf "ProcessJobCommand, Map %A to peer %d" queuePeer.RemoteEndPoint peeri ))
-        try
+    member internal x.ProcessJobCommand( jobAction:SingleJobActionContainer, queuePeer:NetworkCommandQueue, cmd:ControllerCommand, ms:StreamBase<byte>, jobID: Guid ) = 
+//        using ( SingleJobActionContainer.TryFind(jobID)) ( fun jobAction ->
```
zhuweirong: remove the dead code?

zhuweirong:
The UTs are not added yet?

src/tools/tools/LoggerFSharp.fs
```diff
@@ -50,6 +50,18 @@ type Logger private() =
...
+    /// "messageFunc" is evaluated only if "logLevel" is less than or equal to the log level for "logId"
+    static member inline LogF(jobID : Guid, logLevel : LogLevel,messageFunc : unit -> string) =
+        if logLevel <= Logger.DefaultLogLevel then
+            Logger.LoggerProvider.Log((null,jobID, logLevel, messageFunc()))
+
+    /// Log message generated by "messageFunc" using "JobId", 
+    /// "messageFunc" is evaluated only if "logLevel" is less than or equal to the log level for "logId"
+    static member inline LogF(logId: string, jobID : Guid, logLevel : LogLevel,messageFunc : unit -> string) =
+         if Logger.LoggerProvider.IsEnabled(logId, logLevel) then
+            Logger.LoggerProvider.Log((null,jobID, logLevel, messageFunc()))
```
zhuweirong: null --> logId

src/tools/tools/Logger.fs
```diff
@@ -82,6 +83,9 @@ type ILoggerProvider =
     /// Takes three parameters: log id, log level, log Message
     abstract member Log : (string*LogLevel*string) -> unit
 
+    /// Takes four parameters: log id, log level, JobID, log Message
```
zhuweirong: comment does not match code

src/tools/tools/Logger.fs
```diff
@@ -82,6 +83,9 @@ type ILoggerProvider =
...
+    abstract member Log : (string*Guid*LogLevel*string) -> unit
+
```
zhuweirong, line 91: To be consistent, should have an overload that  does (Guid \* LogLevel \* String) -> unit

src/tools/tools/LoggerCSharp.fs
```diff
@@ -47,6 +47,18 @@ type Logger private() =
...
+        if logLevel <= Logger.DefaultLogLevel then
+            Logger.LoggerProvider.Log((null,jobID, logLevel, messageFunc.Invoke()))
+
+    /// Log message generated by "messageFunc" using "JobId", 
+    /// "messageFunc" is evaluated only if "logLevel" is less than or equal to the log level for "logId"
+    static member inline LogF(logId: string, jobID : Guid, logLevel : LogLevel,messageFunc : Func<string>) =
+         if Logger.LoggerProvider.IsEnabled(logId, logLevel) then
+            Logger.LoggerProvider.Log((null,jobID, logLevel, messageFunc.Invoke()))
```
zhuweirong: null -> logId

src/tools/tools/LoggerCSharp.fs
```diff
@@ -47,6 +47,18 @@ type Logger private() =
         if Logger.LoggerProvider.IsEnabled(logId, logLevel) then
             Logger.LoggerProvider.Log((logId, logLevel, messageFunc.Invoke()))
 
+    /// Log message generated by "messageFunc" using "JobId", 
+    /// "messageFunc" is evaluated only if "logLevel" is less than or equal to the log level for "logId"
+    static member inline LogF(jobID : Guid, logLevel : LogLevel,messageFunc : Func<string>) =
+        if logLevel <= Logger.DefaultLogLevel then
+            Logger.LoggerProvider.Log((null,jobID, logLevel, messageFunc.Invoke()))
```
zhuweirong: Once we have an overload of  "(Guid \* LogLevel \* String) -> unit", then there is no need to pass "null" for the first parameter.

src/CoreLib/DSeq.fs
```diff
@@ -532,71 +537,88 @@ and [<AllowNullLiteral>]
...
+            let peeri = x.Cluster.SearchForEndPoint( queuePeer )
+            Logger.LogF( jobID, LogLevel.WildVerbose, ( fun _ -> sprintf "ProcessJobCommand, Map %A to peer %d" queuePeer.RemoteEndPoint peeri ))
             // Outgoing queue for the reply 
             match ( cmd.Verb, cmd.Noun ) with 
             | ( ControllerVerb.Acknowledge, ControllerNoun.DStream ) ->
                 true
             | ( ControllerVerb.SyncWrite, ControllerNoun.DStream ) ->
                 let meta = BlobMetadata.Unpack( ms )
-                Logger.LogF( LogLevel.WildVerbose, ( fun _ -> sprintf "Rcvd %A command %A from peer %d with %dB" (meta.ToString()) cmd peeri (ms.Length-ms.Position) ))
+                Logger.LogF( jobID, LogLevel.MildVerbose, ( fun _ -> sprintf "Rcvd %A command %A from peer %d with %dB" (meta.ToString()) cmd peeri (ms.Length-ms.Position) ))
```
zhuweirong: "Rcvd %A" --> "Rcvd %s", since meta.ToString() returns a string already.

src/CoreLib/DSeq.fs
```diff
@@ -532,71 +537,88 @@ and [<AllowNullLiteral>]
...
                 let bRet = x.SyncReceiveFromPeer meta ms peeri 
 //                let ev = new ManualResetEvent(false)
 //                ev.Reset() |> ignore
 //                ev.WaitOne() |> ignore
                 if bRet && (DateTime.UtcNow - queuePeer.LastSendTicks).TotalMilliseconds > 5000. then 
 //                if bRet then 
-                    let msEcho = new MemStream( 128 )
+                    use msEcho = new MemStream( 128 )
```
zhuweirong, line 559: How the magic size "128" was determined?

src/CoreLib/DSeq.fs
```diff
@@ -532,71 +537,88 @@ and [<AllowNullLiteral>]
...
-                x.bSentPeerCloseConfirmed.[peeri] <- true
-                Logger.Log( LogLevel.MildVerbose, ( sprintf "DStream %s:%s, ConfirmClose received from peer %d" x.Name x.VersionString peeri )    )
-                x.CheckAllPeerClosed()
-            | _ ->
-                Logger.Log( LogLevel.Info, ( sprintf "DStream.Callback(%s:%s), Unexpected command %A, peer %d" x.Name x.VersionString cmd peeri )    )
-        with 
-        | e -> 
-            Logger.Log( LogLevel.Info, ( sprintf "Error in processing DStream.Callback(%s:%s) cmd %A, peer %d, with exception %A" x.Name x.VersionString cmd peeri e )    )
-        true
+//                with
```
zhuweirong: remove dead code.

src/CoreLib/DSeq.fs
```diff
@@ -532,71 +537,88 @@ and [<AllowNullLiteral>]
...
-            Logger.Log( LogLevel.Info, ( sprintf "Error in processing DStream.Callback(%s:%s) cmd %A, peer %d, with exception %A" x.Name x.VersionString cmd peeri e )    )
-        true
+//                with 
+//                | ex -> 
+//                    jobAction.EncounterExceptionAtContainer( ex, sprintf "Error in processing DStream.ProcessJobCommand cmd %A, with payload of %dB" 
+//                                                                        cmd ms.Length )
+//                    true
+//        )
+    /// Parse echoing from DStream from a container to other container
+    member internal x.DStreamCallback( cmd, peeri, ms, jobID, name, verNumber ) =
```
zhuweirong: DStream type is "internal" now. There is no need to mark the member "internal".

src/CoreLib/DSeq.fs
```diff
@@ -532,71 +537,88 @@ and [<AllowNullLiteral>]
...
+        using ( SingleJobActionContainer.TryFind(jobID)) ( fun jobAction -> 
+            if Utils.IsNull jobAction then 
+                Logger.LogF( jobID, LogLevel.Info, ( fun _ -> sprintf "[May be Ok, Job cancelled?] DStream.DStreamCallback, received command %A of payload %dB, but can't find job lifecycle object"
```
zhuweirong: "Info" may get us a lot of log entries here.  In case of a DSet with many elements, and user only does a take 10.  It may take a while for all peers realize the cancellation. At the same time, we may get a lot of such log entries?

src/CoreLib/DSeq.fs
```diff
@@ -673,23 +695,23 @@ and [<AllowNullLiteral>]
...
                 msClose.WriteInt64( x.Version.Ticks )
                 peerQueue.ToSend( ControllerCommand( ControllerVerb.SyncClose, ControllerNoun.DStream), msClose )
-                Logger.Log( LogLevel.MildVerbose, ( sprintf "DStream %s:%s, SyncClose, DStream send to peer %d" x.Name x.VersionString peeri )    )
-                msClose.DecRef()
+                Logger.LogF( jbInfo.JobID, LogLevel.MildVerbose, ( fun _ -> sprintf "DStream %s:%s, SyncClose, DStream send to peer %d" x.Name x.VersionString peeri )    )
             else
                 let msg = sprintf "DStream.SyncClose, DStream %s:%s, attempt to send SyncClose, DStream to peer %d, but the peer queue has already been shutdown" 
                             x.Name x.VersionString peeri 
-                Logger.Log( LogLevel.Warning, msg )
+                Logger.LogF( jbInfo.JobID, LogLevel.Warning, fun _ -> msg )
```
zhuweirong: Not much benefit using LogF here.

Need to add a Log overload that takes JobID.

src/CoreLib/DSeq.fs
```diff
@@ -809,20 +834,20 @@ and [<AllowNullLiteral>]
                         let msg = sprintf "DStream.CloseFromPeer %s:%s has WaitForAllPeerClose flag, it should have a AllPeerCloseRcvdEvent to be flagged on" x.Name x.VersionString
                         Logger.Log( LogLevel.Error, msg )
                         let jbInfo = x.DefaultJobInfo
-                        let msError = new MemStream( 1024 )
+                        use msError = new MemStream( 1024 )
```
zhuweirong, line 828: How the magic size "1024" was determined?

src/CoreLib/DSetAction.fs
```diff
@@ -71,26 +75,38 @@ type internal DSetAction() =
                         for p in parameters do
                             x.Param <- p
     /// Hold a Job object
-    member val Job  = null with get, set
+    member x.Job with get() = jobInternal
+                 and set(v: Job) = Volatile.Write( jobLifecycleRef, JobLifeCycleCollectionApp.BeginJob() )
```
zhuweirong: why "Volatile.Write"?

src/CoreLib/DSetPeer.fs
```diff
@@ -1013,12 +1025,16 @@ and [<AllowNullLiteral>]
                                         yield (sprintf "    Peer %d: (In replicate:%A, ReplicateClose sent: %A, rcvd %A)" i bPeerInReplicate.[i] bCloseDSetSent.[i] bConfirmCloseRcvd.[i] )                                
                             }
                             |> String.concat( System.Environment.NewLine )
-                        let msInfo = new MemStream( 10240 )
-                        msInfo.WriteString( msgInfo ) 
-                        x.HostQueue.ToSend( ControllerCommand( ControllerVerb.Warning, ControllerNoun.Message ) , msInfo )
+                        use msSend = new MemStream( 10240 )
```
zhuweirong: why 10240?

src/CoreLib/GV.fs
```diff
@@ -360,16 +361,520 @@ and /// Information of Within Job cluster information.
...
+type internal JobLifeCycle(jobID:Guid) = 
+    let CTSSource = new CancellationTokenSource()
+    let nExecutorOfDisposeJob = ref 0 
+    let nCancellationCalled = ref 0 
+    let onDispose = ExecuteUponOnce()
+    let onCancellation = ExecuteUponOnce()
+    let onException = ConcurrentQueue<Action<Exception>>()
+    let numExceptionCallback = ref 0
+    let exCollections = ConcurrentQueue<Exception>()
+    static member val MaxWaitToEndJobInMilliseconds = 30000 with get, set
```
zhuweirong: should it be made configurable in deploy.fs?

src/CoreLib/GV.fs
```diff
@@ -360,16 +361,520 @@ and /// Information of Within Job cluster information.
...
+    let nCancellationCalled = ref 0 
+    let onDispose = ExecuteUponOnce()
+    let onCancellation = ExecuteUponOnce()
+    let onException = ConcurrentQueue<Action<Exception>>()
+    let numExceptionCallback = ref 0
+    let exCollections = ConcurrentQueue<Exception>()
+    static member val MaxWaitToEndJobInMilliseconds = 30000 with get, set 
+    member val numJobActionsInProcess = ref 0 with get
+    /// Exception to throw here
+    member x.Exception with get() = if exCollections.IsEmpty then null else AggregateException( exCollections )
```
zhuweirong, line 379: for internal types, we should consider using "option" instead of null.

src/tools/tools/LoggerCSharp.fs
jinlmsft, line 52: Agree. And Will remove inline.

src/tools/tools/Logger.fs
jinlmsft, line 328: Revised. I do have a concern as the ILoggerProvider has 4 Log flavors that needed to be extended when someone implements a customized Log function. Will that be fine?

src/tools/tools/extension.fs
jinlmsft, line 364: For exception serialization (catch remote exception thrown), let's use BinaryFormatter(). The reason is that if the formatter Bruno wrote causes any issue, its exception can be captured and sent back in the Exception framework.

src/tools/tools/extension.fs
jinlmsft, line 370: Reason above to use BinaryFormatter.

src/CoreLib/DSeq.fs
jinlmsft, line 96:
```
/// A collection of DStream that serves as the network interface for intercommunication between Jobs. 
/// The entry is indexed by jobID, while each entry holds a timestamp (in ticks of int64), of the recent input from that job. 
/// That way, we can detect inactive jobs and cancel them later. 
```

src/CoreLib/DSeq.fs
jinlmsft: Removed.

src/tools/tools/LoggerFSharp.fs
jinlmsft: Revised.

src/tools/tools/Logger.fs
jinlmsft, line 91: Done.

src/tools/tools/LoggerCSharp.fs
jinlmsft: Done.

src/tools/tools/LoggerCSharp.fs
jinlmsft: Done.

src/CoreLib/DSeq.fs
jinlmsft: Done.

src/CoreLib/DSeq.fs
jinlmsft, line 559: An estimation of the size of the message. With the change Sanjeev will introduce, I assume that all new MemStream of a size small enough will be equivalent.

src/CoreLib/DSeq.fs
jinlmsft: Internal removed.

src/CoreLib/DSeq.fs
jinlmsft: Good observation. I will change to MildVerbose.

src/CoreLib/DSeq.fs
jinlmsft: Change back to Log.

src/CoreLib/DSetAction.fs
jinlmsft: Agree. Since each DSetAction only host a single job, I will change all Volatile.Write and .Read to normal reference access.

src/CoreLib/DSetPeer.fs
jinlmsft: use 1024.

src/CoreLib/GV.fs
jinlmsft: Agree, and add an entry in deploy.fs.

src/CoreLib/GV.fs
jinlmsft, line 379: Notice that JobLifeCycle and SingleJobAction are both disposable object. You can't use 
use a = new TypeA()

if TypeA is an option.

src/tools/tools/Logger.fs
zhuweirong, line 328: I don't expecting frequent extension though.

tests/CoreLib/DSet.fs
```diff
@@ -1536,3 +1536,70 @@ type DSetTests () =
...
+            for exIn in ex.InnerExceptions do 
+                match exIn with 
+                | :? System.OperationCanceledException as exInner -> 
+                    if String.Compare( exInner.Message, msg, StringComparison.Ordinal) <> 0 then 
+                        bExceptionCatched := false
+                | _ -> 
+                    bExceptionCatched := false
+        | _ -> 
+            ()
+        Assert.AreEqual( !bExceptionCatched, true )
```
zhuweirong: When this assertion fails, it's unclear it's because the expected exception is not caught, or it's because the msg does not match.

PR #67, opened 2015-10-02
Replaced "Everyone" string with the correct security identifier

v-jacques:
The "Everyone" string doesn't get the correct user group on non-English Windows installations.

http://stackoverflow.com/questions/4614697/how-to-get-the-identityreference-for-everyone-to-create-mutexaccessrule-on-loc

https://msdn.microsoft.com/en-us/library/system.security.principal.wellknownsidtype(v=vs.110).aspx


zhuweirong:
Thanks @vijacques!

PR #68, opened 2015-10-03
Adding Matrix-Matrix sample, updating Matrix-Vector

BrunoBozza:


samples/examples/FSharpExamples/Matrix.fs
```diff
@@ -28,38 +28,53 @@ open Prajna.Api.FSharp
...
+        let matrix =  randomMatrix numRows vectorSize
+        let vector = randomVector vectorSize
 
         let singleCoreResult = 
             printTiming "single core" <| lazy 
-                randomMatrix |> Array.map (fun row -> row .* vector)
-        
+                matrix |> Array.map (fun row -> row .* vector)
+
+        let numPartitions = Environment.ProcessorCount
```
zhuweirong, line 65: Why use local machine's processor count to determine the number of partitions for each remote node?

samples/examples/FSharpExamples/Matrix.fs
```diff
@@ -28,38 +28,53 @@ open Prajna.Api.FSharp
...
+        let dSet =
+            printTiming "distributing" <| lazy
+                let d =
+                    DSet<float[]>(Cluster = cluster, NumParallelExecution = numPartitions)
+                    |> DSet.distributeN numPartitions matrix
```
zhuweirong, line 70: I think to make it cached in memory, you need to chain a "|> DSet.cacheInMemory" after distributeN.

samples/examples/FSharpExamples/Matrix.fs
BrunoBozza, line 65: The sample is meant to run locally. Ideally, it would be Environment.ProcessorCount / cluster.NumNodes, but this keeps things simple without hurting speed too much. I should add a comment that on an actual cluster, we would do something different.

zhuweirong, line 65: Note: the example driver does support remote cluster via "-cluster" argument. Only when it was not given, it uses local cluster. I actually use remote clusters to run the examples sometimes.

samples/examples/FSharpExamples/Matrix.fs
BrunoBozza, line 70: I tried, but it doesn't have the same effect, at least not for now, and it was seeming to things worse, and using fold does what I wanted. To be sure, after your comment (maybe I was imagining things), I ran an experiment: First, to set a baseline, I ran 5 times without calling cacheInMemory or fold. Here are the results:

single core: 4.107
distributing: 0.004
computing distributed: 1.407

single core: 4.806
distributing: 0.004
computing distributed: 1.328

single core: 4.556
distributing: 0.004
computing distributed: 1.325

single core: 4.667
distributing: 0.004
computing distributed: 1.351

single core: 3.924
distributing: 0.004
computing distributed: 1.356

Next I used cacheInMemory. Eventually, we should make this work well, perhaps under a different name. But for now, here's what I get when I add a "|> DSet.cacheInMemory":

single core: 4.711
distributing: 0.028
computing distributed: 1.487

single core: 4.568
distributing: 0.012
computing distributed: 1.465

single core: 4.62
distributing: 0.013
computing distributed: 1.382

single core: 4.096
distributing: 0.013
computing distributed: 1.399

single core: 4.075
distributing: 0.013
computing distributed: 1.425

Finally, to make sure the fold was actually doing something, I ran 5 times using fold, as in the current code. Here's what I get:

single core: 4.03
distributing: 0.634
computing distributed: 1.061

single core: 4.476
distributing: 0.63
computing distributed: 1.162

single core: 4.158
distributing: 0.625
computing distributed: 1.138

single core: 4.783
distributing: 0.706
computing distributed: 1.087

single core: 4.787
distributing: 0.588
computing distributed: 1.145

So adding a cacheInMemory instead of fold makes things ~30%-50% slower when running distributed. We should fix this, of course, but the purpose of the sample is to demonstrate how you can use Prajna to do a distributed Matrix-Matrix multiply with minimum overhead, and the straightforward way to do this for now is adding a dummy fold before the main computation.

samples/examples/FSharpExamples/Matrix.fs
BrunoBozza, line 65: Ok. What about something simple like cluster.NumNodes \* 100? Alternatively, we could run a precomputation on all nodes to extract the machine name and number of processors on each node, then add the nodes per unique machine. But I don't think our APIs even guarantee distribution, so we would have to either add that guarantee to an API (maybe DSet.distribute?) or go lower level to extract the ideal counts. Any preference? Other options?

(Update: just changed suggestion to "\* 100". It doesn't have to be exact, just something that's more processors than most cluster nodes, but doesn't introduce too much overhead)

samples/examples/FSharpExamples/Matrix.fs
zhuweirong, line 70: I guess I didn't make myself clear.  First, the "fold" on "d" will not cause "d" to be persisted to memory. When you use "d" in other places, the operations that computes "d" (e.g. distributeN) will be re-evaluated. So you don't achieve what you want. (If it somehow does, then something is wrong in your measurement).  Secondly, "cacheInMemory" itself is not an action thus it will not be evaluated unless there is an action following it.  That's why itself alone does not help either. 

What you need is

```
    let dSet =
        printTiming "distributing" <| lazy
            let d =
                DSet<float[]>(Cluster = cluster, NumParallelExecution = numPartitions)
                |> DSet.distributeN numPartitions matrix
                |> DSet.cacheInMemory
            // Force push DSet to cluster
            do d |> DSet.fold (fun _ _ -> 0) (fun _ _ -> 0) 0 |> ignore
            d
```

That is, you use DSet.fold to trigger it to be cached. Thus the returned "d" is the cached one. When it was used in other computation graph, it will not need to be re-evaluated from the root. 

This is how it is supposed to work now. If you see anything unexpected, then there're something wrong.  But this behavior was verified in UT.  See https://github.com/BrunoBozza/Prajna/blob/master/tests/CoreLib/DSet.fs#L1289, you can use the same trick as the test to verify the behavior.

samples/examples/FSharpExamples/Matrix.fs
zhuweirong, line 65: It can be thought in a different way. A partition is a unit for scheduling and failure recovery. The proper size of a partition can often be determined by the application itself instead of # of nodes. Usually, # of partitions can be larger or even far larger than # of nodes (or total # of processors).

BrunoBozza, line 65: So, what would be a good number for this sample, which I think we can think of as application itself? We have (almost) perfectly even computation to be distributed between the nodes, with essentially no variance in compute time.

samples/examples/FSharpExamples/Matrix.fs
BrunoBozza, line 70: Oh... That I have nothing against. I have seen some instability in run time with cacheInMemory, though, but I can't repro now, so I think it's good.

zhuweirong, line 70: The point is "fold" itself alone would/should not achieve the pre-loading effect.

PR #74, opened 2015-10-15
Distributed Function (local part).

jinlmsft:
This series of changes merge previous distributed function supported in Visual Media Hub (VM Hub) into Prajna Core.
The idea is to implement a distributed function, which can be considered as .Net Remoting or WCF, but instead of
calling and executing a function on a remote peer, Prajna distributed function can be executed against a cluster of node for
varying functionality (e.g., choose one for executing a distributed service, or choose N node to execute a Paxos consesus protocol).

Each distributed function is indentified with a four tuple: provider, domain, schemaIn, schemaOut. It can be considered as a
function delegate Func<T,TResult> to be executed in a cluster.
- SchemaIn is a Guid indentify the data type and coding of type T.
- SchemaOut is a Guid identify the data type and coding of type TResult.
- Domain is a Guid that identify the name of the function (as there can be multiple function with the same schema). It can be
  converted from a name (string).
- Provider identifies the operator and version of the function.

The service endpoint register a distributed function through:
- RegisterUnitAction( name, unit -> unit )
- RegisterFunction<_,_>( name, T -> TResult )

The registered function return a IDisposable interface, and when disposed, the registered function will not be further used.
The service provider may specify a concurrency level of the function by setting:

DistributedFunctionStore.Current.ConcurrentCapacity
- When the concurrency level is 1 (default), the function is invoked sequentially. Only one user may enter each function at a time.
  This should always be used when the application supported by the function is not multi-thread safe.
- When the concurrency level is N (N>1), the system allows N instance of the function to be concurrently called each time. Additional
  caller will have to wait in queue.
- When the concurrency level is negative or 0, no concurrency rule is assumed. The system assume that the function is lock-free, and may
  be executed with any degree of freedom.
  Note that the concurrency level of Int32.MaxValue and 0 may lead to different performance. The former still have one Interlocked.Increment
  and Interlocked.Decrement function, which may slow down function execution at very high concurrency level.

To use the distributed function, the App will import the distributed function as:
- TryImportUnitActionLocal( name )
- TryImportFunction<_,_>( name )

It can then simply execute the function/action.

This check-in only include the local operation of the distributed function and unit test. The remoting portion of the distributed function is still
in the work.

### DIFF
--- a/src/tools/tools/process.fs
+++ b/src/tools/tools/process.fs
@@ -153,6 +153,8 @@ type internal OneCleanUp ( o:Object, infoFunc, f, earlyCleanUp: unit -> unit ) =
 
 // In Prajna, the order of clean up is 
 // ThreadPoolTimerCollection:       300
+// ContainerJobCollection:          500
+// DistributedFunctionStore:        700
 // Cluster:                   1000
 // ThreadPoolWithWaitHandles:       1500
 // ThreadPoolWait:                  2000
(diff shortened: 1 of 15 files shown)


src/tools/tools/process.fs
```diff
@@ -1343,6 +1345,17 @@ and internal ThreadPoolWaitHandles() =
             x.WaitOne(ev, millisecondsTimeout)
         else
             bStatus
+    member private x.WaitAny(ev, millisecondsTimeout:int) =
```
zhuweirong, line 1348: "ev" is an array of handles, right? Better to use plural.

PR #75, opened 2015-10-16
Fixes for ContainerAppDomainInfo and serialization perf regression, and new PrajnaFormatter schema

BrunoBozza:
This pull request has 3 independent commits that I'd like to keep, since they address different issues.
1) Only today I noticed a major perf regression on closure serialization. A simple test serializaing a closure carrying a few megabytes shows a > 3x difference.
2) AppDomain cluster was broken if using external assemblies. Somehow it was resolving user assemblies, but if those refer to others, that was not getting loaded.
3) Adding SchemaPrajnaFormatterHelper per Jin's request, in case we send more than a few kilobytes to a distributed function.

### DIFF
--- a/src/CoreLib/distributedfunction.fs
+++ b/src/CoreLib/distributedfunction.fs
@@ -205,10 +205,27 @@ type internal SchemaBinaryFormatterHelper<'T>() =
         let o = fmt.Deserialize( ms ) 
         if Utils.IsNotNull o then o :?> 'T else Unchecked.defaultof<_>
 
+/// Helper class to generate schema ID
+type internal SchemaPrajnaFormatterHelper<'T>() = 
+    /// Get a GUID that representing the coding of a type. Note that this binds to the 
+    /// fullname of the type with assembly (AssemblyQualifiedName) so if the assembly 
+    /// of the type changes, the schemaID will change. 
+    static member SchemaID() = 
+        let schemaStr = "Prajna.Tools.BinarySerializer:" + typeof<'T>.AssemblyQualifiedName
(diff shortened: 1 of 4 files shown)


src/CoreLib/distributedfunction.fs
```diff
@@ -205,10 +205,27 @@ type internal SchemaBinaryFormatterHelper<'T>() =
         let o = fmt.Deserialize( ms ) 
         if Utils.IsNotNull o then o :?> 'T else Unchecked.defaultof<_>
 
+/// Helper class to generate schema ID
+type internal SchemaPrajnaFormatterHelper<'T>() =
```
zhuweirong, line 209: We'd prefer not having "Prajna" in type/member name but only in namespace, especially for public visible ones. This is internal, so it's not a big deal, just point it out.

src/CoreLib/distributedfunction.fs
```diff
@@ -205,10 +205,27 @@ type internal SchemaBinaryFormatterHelper<'T>() =
...
+    static member Decoder(ms:Stream) = 
+        let fmt = Prajna.Tools.BinarySerializer() :> IFormatter
+        let o = fmt.Deserialize( ms ) 
+        if Utils.IsNotNull o then o :?> 'T else Unchecked.defaultof<_>
+
 /// Govern the behavior of the default serialization to be used 
 type DefaultSerializerForDistributedFunction =
     /// Default Serializer implementation in Prajna
-    | DefaultSerializer = 0
+    | PrajnaSerializer = 0
```
zhuweirong, line 228: Same as above.

PR #76, opened 2015-10-16
Alternative implementation of ThreadPoolWait.

jinlmsft:
ThreadPoolWait implement functionality of ThreadPoolWait .RegisterWaitForSingleObject, with three capability:
1. To remove a particular wait handle and cancel its operation
2. To show all wait handles queued in the system,
   and 3. To show long continuation (not normal).

The old implementation uses a customized threadpool. The new implementation uses ThreadPoolWait.RegisterWaitForSingleObject.

One potential performance issue is that:

ThreadPoolWait.MaxSpinMs = 200 ms.

This is the time between the callback is called, and the registeredWaitHandle is put into the system. I found that if I set

ThreadPoolWait.MaxSpinMs = 50 ms.

Sometime, the system will show it is unable to grab the registeredWaitHandle in 50ms. I can't reason why it takes so long for the
ThreadPoolWait.RegisterWaitForSingleObject to get back. In slower system, we may need to use a even larger wait.

### DIFF
--- a/src/tools/tools/process.fs
+++ b/src/tools/tools/process.fs
@@ -873,9 +873,9 @@ type internal ExecutionTasks =
 /// Common portion of the customized threadpool, wait for handles. 
 /// One wait thread will be spinned every 64 handles
 /// </summary>
-type internal ThreadPoolWait internal (id:int) as this = 
+type private ThreadPoolWaitDeprecated internal (id:int) as this = 
     do 
-        CleanUp.Current.Register( 2000, this, ThreadPoolWait.TerminateAll, fun _ -> "ThreadPoolWait" ) |> ignore 
+        CleanUp.Current.Register( 2000, this, ThreadPoolWaitDeprecated.TerminateAll, fun _ -> "ThreadPoolWaitDeprecated" ) |> ignore 
     // Each thread can wait at most 64 handles, 
     static member val private MAX_WAITHANDLES = 64 with get, set
     /// bTerminate: stop all waiting threads. 
@@ -884,31 +884,31 @@ type internal ThreadPoolWait internal (id:int) as this =
     static member val private evTerminate = new ManualResetEvent(false) with get
     static member val private ActiveThreadPools = ConcurrentDictionary<_,_>() with get
     static member RegisterThreadPool (name:string) = 
-        Logger.LogF( LogLevel.MediumVerbose, ( fun _ -> sprintf "ThreadPoolWait, register wait pool %s ............" name ))
-        while (!ThreadPoolWait.nTerminate)<>0 do
+        Logger.LogF( LogLevel.MediumVerbose, ( fun _ -> sprintf "ThreadPoolWaitDeprecated, register wait pool %s ............" name ))
+        while (!ThreadPoolWaitDeprecated.nTerminate)<>0 do
             // Old Waiting Threads are being terminated, need to wait for that procedure to end. 
             let spin = SpinWait()
             spin.SpinOnce()             
-        ThreadPoolWait.ActiveThreadPools.Item( name ) <- (PerfADateTime.UtcNow())
+        ThreadPoolWaitDeprecated.ActiveThreadPools.Item( name ) <- (PerfADateTime.UtcNow())
     static member UnregisterThreadPool (name:string) = 
-        Logger.LogF( LogLevel.MildVerbose, ( fun _ -> sprintf "ThreadPoolWait, unregister wait pool %s ............" name ))
-        ThreadPoolWait.ActiveThreadPools.TryRemove( name ) |> ignore
-        if ThreadPoolWait.ActiveThreadPools.IsEmpty then 
-            if ( !(ThreadPoolWait.NumberOfWaitingThreads) > 0 ) then 
-                Logger.LogF( LogLevel.MildVerbose, ( fun _ -> sprintf "ThreadPoolWait, threadpool %s is the last registered threadpools, try terminating all" name ))
+        Logger.LogF( LogLevel.MildVerbose, ( fun _ -> sprintf "ThreadPoolWaitDeprecated, unregister wait pool %s ............" name ))
+        ThreadPoolWaitDeprecated.ActiveThreadPools.TryRemove( name ) |> ignore
+        if ThreadPoolWaitDeprecated.ActiveThreadPools.IsEmpty then 
+            if ( !(ThreadPoolWaitDeprecated.NumberOfWaitingThreads) > 0 ) then 
+                Logger.LogF( LogLevel.MildVerbose, ( fun _ -> sprintf "ThreadPoolWaitDeprecated, threadpool %s is the last registered threadpools, try terminating all" name ))
             // Termniate all Wait thread. 
-            ThreadPoolWait.TryTerminateAll()
-    static member val WaitingThreads = ConcurrentQueue<ThreadPoolWait*Thread>() with get, set
+            ThreadPoolWaitDeprecated.TryTerminateAll()
+    static member val WaitingThreads = ConcurrentQueue<ThreadPoolWaitDeprecated*Thread>() with get, set
     static member val NumberOfWaitingThreads = ref 0 with get, set
     static member private ClearAll() = 
         // Clear Up Structure. 
         let refValue = ref Unchecked.defaultof<_>
-        while ThreadPoolWait.WaitingThreads.TryDequeue( refValue ) do 
+        while ThreadPoolWaitDeprecated.WaitingThreads.TryDequeue( refValue ) do 
             let tp, _ = !refValue;
             (tp :> IDisposable).Dispose()
-        ThreadPoolWait.nTerminate := 0
+        ThreadPoolWaitDeprecated.nTerminate := 0
     /// <summary>
-    /// ThreadPoolWait.WaitForHandle schedule a continuation function to be executed when handle fires. 
+    /// ThreadPoolWaitDeprecated.WaitForHandle schedule a continuation function to be executed when handle fires. 
     /// </summary>
     /// <param name="infoFunc"> Information delegate of the handle/continution to be waited for, used in diagnostics </param>
     /// <param name="handle"> The handle to be waited on. </param>
@@ -917,56 +917,56 @@ type internal ThreadPoolWait internal (id:int) as this =
     /// </param> 
     /// <param name="unblockHandle"> handle to set if continuation function fired. </param>
     static member WaitForHandle (infoFunc:unit->string) (handle:WaitHandle) continuation (unblockHandle:ManualResetEvent) = 
-        if (!ThreadPoolWait.nTerminate)<>0 then 
-            let msg = sprintf "ThreadPoolWait.WaitForHandle should not be called when the corresponding Threadpool has been unregistered!" 
+        if (!ThreadPoolWaitDeprecated.nTerminate)<>0 then 
+            let msg = sprintf "ThreadPoolWaitDeprecated.WaitForHandle should not be called when the corresponding Threadpool has been unregistered!" 
             Logger.Log( LogLevel.Error, msg )
             failwith msg 
         else
             let mutable bDone = false
             while not bDone do 
-                for tuple in ThreadPoolWait.WaitingThreads do 
+                for tuple in ThreadPoolWaitDeprecated.WaitingThreads do 
                     let pool, _ = tuple
                     if not bDone && not pool.IsFull && pool.EnqueueWaitHandle infoFunc handle continuation unblockHandle then 
                         bDone <- true
                 if not bDone then 
-                    // Add a new ThreadPoolWait object, we don't use Interlocked.Increment to avoid creating multiple waiting threads. 
-                    lock ( ThreadPoolWait.WaitingThreads ) ( fun _ -> 
+                    // Add a new ThreadPoolWaitDeprecated object, we don't use Interlocked.Increment to avoid creating multiple waiting threads. 
+                    lock ( ThreadPoolWaitDeprecated.WaitingThreads ) ( fun _ -> 
                             // For the thread that enters first, this gets executed. 
                             // Starting the waiting thread. 
-                            let threadid = Interlocked.Increment( ThreadPoolWait.NumberOfWaitingThreads ) - 1
-                            let curPool = new ThreadPoolWait(threadid)
-                            let thread = ThreadTracking.StartThreadForFunction ( fun _ -> sprintf "ThreadPoolWait waiting thread %d" threadid) (curPool.Wait)
-                            ThreadPoolWait.WaitingThreads.Enqueue( (curPool, thread) )  
-                            Logger.LogF( LogLevel.MildVerbose, ( fun _ -> sprintf "ThreadPoolWait, launching waiting thread %d by %s" threadid (infoFunc()) ))
+                            let threadid = Interlocked.Increment( ThreadPoolWaitDeprecated.NumberOfWaitingThreads ) - 1
+                            let curPool = new ThreadPoolWaitDeprecated(threadid)
+                            let thread = ThreadTracking.StartThreadForFunction ( fun _ -> sprintf "ThreadPoolWaitDeprecated waiting thread %d" threadid) (curPool.Wait)
+                            ThreadPoolWaitDeprecated.WaitingThreads.Enqueue( (curPool, thread) )  
+                            Logger.LogF( LogLevel.MildVerbose, ( fun _ -> sprintf "ThreadPoolWaitDeprecated, launching waiting thread %d by %s" threadid (infoFunc()) ))
                             if threadid > 2 then 
-                                ThreadPoolWait.MonitorAllWaitHandles()   
+                                ThreadPoolWaitDeprecated.MonitorAllWaitHandles()   
                         )
         ()
 
     // First of the waiting hanle is a waiting handle used for informing the arriving of new handles. 
     static member private InitWaitingHandles() = 
-        let arr = Array.zeroCreate ThreadPoolWait.MAX_WAITHANDLES
+        let arr = Array.zeroCreate ThreadPoolWaitDeprecated.MAX_WAITHANDLES
         arr.[0] <- new ManualResetEvent(false) :> WaitHandle
         arr 
     // Use List<_> to hold waiting threads, we don't want to create too many Waiting Threads. 
     member val ThreadID = id with get
-    member val WaitingHandles = ThreadPoolWait.InitWaitingHandles() with get
-    member val Continuations = Array.zeroCreate ThreadPoolWait.MAX_WAITHANDLES with get
-    member val UnblockHandles = Array.zeroCreate ThreadPoolWait.MAX_WAITHANDLES with get
-    member val InfoFunc = Array.zeroCreate ThreadPoolWait.MAX_WAITHANDLES with get
+    member val WaitingHandles = ThreadPoolWaitDeprecated.InitWaitingHandles() with get
+    member val Continuations = Array.zeroCreate ThreadPoolWaitDeprecated.MAX_WAITHANDLES with get
+    member val UnblockHandles = Array.zeroCreate ThreadPoolWaitDeprecated.MAX_WAITHANDLES with get
+    member val InfoFunc = Array.zeroCreate ThreadPoolWaitDeprecated.MAX_WAITHANDLES with get
     member val LastWaitingHandles = ref 0 with get, set
     /// Information to detect deadlock in cont() 
     /// i.e., a blocking operation in cont()
     member val ContinueTicks = -1L with get, set
     member val ContinueInfo = Unchecked.defaultof<_> with get, set
     member val private Lock = SpinLockSlim(true) with get, set
-    member x.IsFull with get() = (!x.LastWaitingHandles) >= ThreadPoolWait.MAX_WAITHANDLES - 1
+    member x.IsFull with get() = (!x.LastWaitingHandles) >= ThreadPoolWaitDeprecated.MAX_WAITHANDLES - 1
     // Multiple thread on queueing. 
     member x.EnqueueWaitHandle (infoFunc:unit->string) handle continuation unblockHandle =
         if not (handle.WaitOne(0)) then 
             x.Lock.Enter()
             let idx = Interlocked.Increment( x.LastWaitingHandles ) 
-            if idx >= ThreadPoolWait.MAX_WAITHANDLES then 
+            if idx >= ThreadPoolWaitDeprecated.MAX_WAITHANDLES then 
                 Interlocked.Decrement( x.LastWaitingHandles ) |> ignore 
                 // We have exhaussted the waiting slot
                 x.Lock.Exit()
@@ -984,7 +984,7 @@ type internal ThreadPoolWait internal (id:int) as this =
                     handle0.Set() |> ignore
                     Logger.LogF( LogLevel.WildVerbose, ( fun _ -> sprintf "add wait handle %s to pos %d, threadID %d " (infoFunc()) idx x.ThreadID ))
                 else
-                    let msg = sprintf "ThreadPoolWait.EnqueueWaitHandle, should never store the handle in slot 0"
+                    let msg = sprintf "ThreadPoolWaitDeprecated.EnqueueWaitHandle, should never store the handle in slot 0"
                     Logger.Log( LogLevel.Error, msg )
                     failwith msg
                 x.Lock.Exit()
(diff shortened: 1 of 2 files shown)


src/tools/tools/process.fs
```diff
@@ -1171,10 +1171,130 @@ type internal ThreadPoolWait internal (id:int) as this =
...
+                    continuation() 
+                    if Utils.IsNotNull unblockHandle  then 
+                        // If there is an unblock handle, set it. 
+                        unblockHandle.Set() |> ignore  
+                    (!handleCollectionRef).TryRemove( continuationObject ) |> ignore 
+                    if (!handleCollectionRef).IsEmpty then 
+                        ThreadPoolWait.WaitHandleCollection.TryRemove( handle) |> ignore
+                    rwh.Unregister(null) |> ignore 
+                else
+                    failwith (sprintf "ThreadPoolWait.Callback, can't find the entry of waithandle and continuation object after %d ms, reason: %s " ThreadPoolWait.MaxSpinMs !failedReason )
```
zhuweirong: This is unrecoverable error. On slow system, it can totally happen. However, there's nothing wrong about it, I think we should not fail here. 

I think if we handle "rwh" as the example in the email, then we can add the item to the handleCollection before calling RegisterWaitForSingleObject.  As a result, there is no need to spin here, it will be a deterministic behavior. 

BTW: I assume the purpose WaitHandleCollection is to help on monitoring and remove?

src/tools/tools/process.fs
```diff
@@ -1171,10 +1171,130 @@ type internal ThreadPoolWait internal (id:int) as this =
...
+            ThreadPoolWait.TryRemove( pair0.Key ) |> ignore 
+        ()
+    /// Mark all functions associated with handle as cancelled, the continuation function will not execute
+    static member TryRemove (handle:WaitHandle) = 
+        let bExist, handleCollection = ThreadPoolWait.WaitHandleCollection.TryGetValue( handle )
+        if bExist then 
+            for pair in handleCollection do 
+                let rwh, infoFunc, workRef, refStartContinuation = pair.Value
+                // If this get executed, the actual work will be cancelled. 
+                Volatile.Write( workRef,  (( fun _ -> () ), null) )
```
zhuweirong: This code seems rather unusual way to achieve cancellation. Also it can be dangerous, for example, at

   let continuation, unblockHandle = !workRef

depends on the F# implementation,  we may get continuation before it was set to null, but "unblockHandle" after it becomes null?  It's totally OK for the compiler to generate code like

   continuation = workRef.Item1
   unblockHandle = workRef.Item2

I'd recommend not doing this. 

Let's take a step back: In what case, we would want to remove a handle without getting cancellation request?  I think the "callback" should take a CancellationToken as a input parameter, and do cooperative cancellation inside it.  We also put the cancellation into the collection. And we should further assume TryRemove will be called only if the cancellation is requested.   Here, we can check whether the cancellation has been requested. If not,w e should regard it as an error and do not try to remove the handle. 

If we really really have the case where we want to remove the WaitHandle but somehow has no CancellationToken from upstream (I really doubt that should be the case),  we can try to create a CancellationTokenSource and add it to collection too (should be null, if we already have a token from upstream).  In this case, we get a token from our own just created source and pass it to then callback. During TryRemove, if we find CancellationTokenSource is not null, then we request the cancel, then set the handle.

src/tools/tools/process.fs
```diff
@@ -1171,10 +1171,130 @@ type internal ThreadPoolWait internal (id:int) as this =
...
+            continuation() 
+            if Utils.IsNotNull unblockHandle then 
+                // If there is an unblock handle, set it. 
+                unblockHandle.Set() |> ignore  
+        else
+            let continuationObject = continuation :> Object
+            let rwh = ThreadPool.RegisterWaitForSingleObject( handle, new WaitOrTimerCallback(ThreadPoolWait.CallBack), (handle,continuationObject) , -1, true )
+            let handleCollection = ThreadPoolWait.WaitHandleCollection.GetOrAdd( handle, fun _ -> ConcurrentDictionary())
+            handleCollection.Item(continuationObject) <- ( rwh, infoFunc, ref (continuation, unblockHandle), ref DateTime.MinValue.Ticks )
+    static member private CallBack( state: Object ) (timeout:bool) =
```
zhuweirong, line 1206: the callback should take cancellation token as a input parameter, so it can do cooperative cancellation.

src/tools/tools/process.fs
```diff
@@ -1171,10 +1171,130 @@ type internal ThreadPoolWait internal (id:int) as this =
...
+        ThreadPoolWait.ActiveThreadPools.Item( name ) <- DateTime.UtcNow
+    static member UnregisterThreadPool (name:string) = 
+        Logger.LogF( LogLevel.MildVerbose, ( fun _ -> sprintf "ThreadPoolWait, unregister wait pool %s ............" name ))
+        ThreadPoolWait.ActiveThreadPools.TryRemove( name ) |> ignore
+        if ThreadPoolWait.ActiveThreadPools.IsEmpty then 
+            if ( !(ThreadPoolWaitDeprecated.NumberOfWaitingThreads) > 0 ) then 
+                Logger.LogF( LogLevel.MildVerbose, ( fun _ -> sprintf "ThreadPoolWait, threadpool %s is the last registered threadpools, try terminating all" name ))
+            // Termniate all Wait thread. 
+            ThreadPoolWait.TryTerminateAll()
+    static member WaitForHandle (infoFunc: unit-> string) (handle:WaitHandle) (continuation:unit->unit) (unblockHandle:ManualResetEvent) =
```
zhuweirong: should take in a cancellation token.

zhuweirong, line 1195: user should also provide a cancellation token (which can be null), thus we can do cooperative cancellation

src/tools/tools/process.fs
```diff
@@ -1171,10 +1171,130 @@ type internal ThreadPoolWait internal (id:int) as this =
...
+        if handle.WaitOne(0) then 
+            continuation() 
+            if Utils.IsNotNull unblockHandle then 
+                // If there is an unblock handle, set it. 
+                unblockHandle.Set() |> ignore  
+        else
+            let continuationObject = continuation :> Object
+            let rwh = ThreadPool.RegisterWaitForSingleObject( handle, new WaitOrTimerCallback(ThreadPoolWait.CallBack), (handle,continuationObject) , -1, true )
+            let handleCollection = ThreadPoolWait.WaitHandleCollection.GetOrAdd( handle, fun _ -> ConcurrentDictionary())
```
zhuweirong: the cancellation token should be put into the collection too, thus can be used later (see comments at TryRemove)

zhuweirong:
Great stuff! I have some comments in the code, please check them out.

src/tools/tools/process.fs
```diff
@@ -1171,10 +1171,146 @@ type internal ThreadPoolWait internal (id:int) as this =
...
+          
+/// ThreadPoolWait.RegisterWaitForSingleObject, with capability:
+/// 1. To remove a particular wait handle. 
+/// 2. To show all wait handles queued in the system, 
+/// and 3. To show long continuation (not normal). 
+type internal ThreadPoolWait() = 
+    static let rwhToRemove = ConcurrentQueue<_>()
+    static let rwhLookup = ConcurrentDictionary<_,_>()
+    /// Maximum wait for the RegisteredWaitHandle to show up
+    static member val private MaxSpinMs = 0 with get, set
```
zhuweirong: MaxSpinMs seems not used anywhere, should be removed.

src/tools/tools/process.fs
```diff
@@ -1171,10 +1171,146 @@ type internal ThreadPoolWait internal (id:int) as this =
...
+    static member val WarningWaitInMs = 10 with get, set
```
zhuweirong: Not used either.

src/tools/tools/process.fs
```diff
@@ -1171,10 +1171,146 @@ type internal ThreadPoolWait internal (id:int) as this =
...
+    static member val ActiveThreadPools = ConcurrentDictionary<_,_>() with get
+    static member val WaitHandleCollection = ConcurrentDictionary<_,ConcurrentDictionary<_,_>>() with get
+    static member RegisterThreadPool (name:string) =
```
zhuweirong: Should ActiveThreadPools/WaitHandleCollections be private?

src/tools/tools/process.fs
```diff
@@ -1171,10 +1171,146 @@ type internal ThreadPoolWait internal (id:int) as this =
...
+/// 2. To show all wait handles queued in the system, 
+/// and 3. To show long continuation (not normal). 
+type internal ThreadPoolWait() = 
+    static let rwhToRemove = ConcurrentQueue<_>()
+    static let rwhLookup = ConcurrentDictionary<_,_>()
+    /// Maximum wait for the RegisteredWaitHandle to show up
+    static member val private MaxSpinMs = 0 with get, set
+    static member val WarningWaitInMs = 10 with get, set
+    static member val ActiveThreadPools = ConcurrentDictionary<_,_>() with get
+    static member val WaitHandleCollection = ConcurrentDictionary<_,ConcurrentDictionary<_,_>>() with get
```
zhuweirong: Should ActiveThreadPools/WaitHandleCollections be private?

src/tools/tools/process.fs
```diff
@@ -1171,10 +1171,146 @@ type internal ThreadPoolWait internal (id:int) as this =
...
+    static member RegisterThreadPool (name:string) = 
+        Logger.LogF( LogLevel.MediumVerbose, ( fun _ -> sprintf "ThreadPoolWait, register wait pool %s ............" name ))
+        ThreadPoolWait.ActiveThreadPools.Item( name ) <- DateTime.UtcNow
+    static member UnregisterThreadPool (name:string) = 
+        Logger.LogF( LogLevel.MildVerbose, ( fun _ -> sprintf "ThreadPoolWait, unregister wait pool %s ............" name ))
+        ThreadPoolWait.ActiveThreadPools.TryRemove( name ) |> ignore
+        if ThreadPoolWait.ActiveThreadPools.IsEmpty then 
+            if ( !(ThreadPoolWaitDeprecated.NumberOfWaitingThreads) > 0 ) then
```
zhuweirong: Why check ThreadPoolWaitDeprecated?

src/tools/tools/process.fs
```diff
@@ -1171,10 +1171,146 @@ type internal ThreadPoolWait internal (id:int) as this =
...
+    /// Maximum wait for the RegisteredWaitHandle to show up
+    static member val private MaxSpinMs = 0 with get, set
+    static member val WarningWaitInMs = 10 with get, set
+    static member val ActiveThreadPools = ConcurrentDictionary<_,_>() with get
+    static member val WaitHandleCollection = ConcurrentDictionary<_,ConcurrentDictionary<_,_>>() with get
+    static member RegisterThreadPool (name:string) = 
+        Logger.LogF( LogLevel.MediumVerbose, ( fun _ -> sprintf "ThreadPoolWait, register wait pool %s ............" name ))
+        ThreadPoolWait.ActiveThreadPools.Item( name ) <- DateTime.UtcNow
+    static member UnregisterThreadPool (name:string) = 
+        Logger.LogF( LogLevel.MildVerbose, ( fun _ -> sprintf "ThreadPoolWait, unregister wait pool %s ............" name ))
```
zhuweirong: When should RegisterThreadPool/UnregisterThreadPool be called? What's the purpose? Or they are here for the purpose of backwards compatibility for existing code? Either way, please add some comments?

src/tools/tools/process.fs
```diff
@@ -1171,10 +1171,146 @@ type internal ThreadPoolWait internal (id:int) as this =
...
+/// 1. To remove a particular wait handle. 
+/// 2. To show all wait handles queued in the system, 
+/// and 3. To show long continuation (not normal). 
+type internal ThreadPoolWait() = 
+    static let rwhToRemove = ConcurrentQueue<_>()
+    static let rwhLookup = ConcurrentDictionary<_,_>()
+    /// Maximum wait for the RegisteredWaitHandle to show up
+    static member val private MaxSpinMs = 0 with get, set
+    static member val WarningWaitInMs = 10 with get, set
+    static member val ActiveThreadPools = ConcurrentDictionary<_,_>() with get
```
zhuweirong: What's the purpose of ActiveThreadPools?

src/tools/tools/process.fs
```diff
@@ -1171,10 +1171,146 @@ type internal ThreadPoolWait internal (id:int) as this =
...
+            continuation() 
+            if Utils.IsNotNull unblockHandle then 
+                // If there is an unblock handle, set it. 
+                unblockHandle.Set() |> ignore
+        else
+            /// Uniquely identify this async job and its resource removal. 
+            let jobObject = Object()
+            let rwh = ThreadPool.RegisterWaitForSingleObject( handle, new WaitOrTimerCallback(ThreadPoolWait.CallBack), (infoFunc,handle,continuation,unblockHandle,jobObject) , -1, true )
+            let handleCollection = ThreadPoolWait.WaitHandleCollection.GetOrAdd( handle, fun _ -> ConcurrentDictionary())
+            handleCollection.Item(jobObject) <- ( rwh, infoFunc, ref DateTime.MinValue.Ticks )
```
zhuweirong: so we assume that more than one WaitCallback can be registered to wait on the same handle, right?

src/tools/tools/process.fs
```diff
@@ -1171,10 +1171,146 @@ type internal ThreadPoolWait internal (id:int) as this =
...
+                    continuation() 
+                    if Utils.IsNotNull unblockHandle  then 
+                        // If there is an unblock handle, set it. 
+                        unblockHandle.Set() |> ignore  
+                if bExist then 
+                    let rwh, _, _ = tuple 
+                    rwh.Unregister(null) |> ignore 
+                    handleCollection.TryRemove( jobObject ) |> ignore 
+                    if handleCollection.IsEmpty then
+                        ThreadPoolWait.WaitHandleCollection.TryRemove( handle ) |> ignore
```
zhuweirong: This is not safe to do. It's possible this thread sees handleCollection.IsEmpty, then takes the branch, however, before it called "TryRemove", another thread can register a wait for the same handle.  This remove would cause the new handle to be removed, which should not happen.

src/tools/tools/process.fs
jinlmsft: Will remove

src/tools/tools/process.fs
jinlmsft: will change to static let.

src/tools/tools/process.fs
jinlmsft: will remove RegisterThreadPool and UnregisterThreadPool.

zhuweirong:
In the test run https://ci.appveyor.com/project/zhuweirong/prajna-rbpft/build/0.0.1.259, though it's successful, it seems the test time has significantly increased (compared to latest merged pull request: https://ci.appveyor.com/project/zhuweirong/prajna-rbpft/build/0.0.1.254). (I'm running it again to see if it's just a glitch).

src/tools/tools/process.fs
jinlmsft: The original thought is to track a pool of handles related to a particular job. But with the job lifecycle object, it may not be necessary, as we can simply register handle wakeup through the CancellationToken with related to the job. I decide to remove both RegisterThreadPool  and UnregisterThreadPool.

src/tools/tools/process.fs
jinlmsft: that is correct. Multiple waitcallback can be registered to wait on a handle.

src/tools/tools/process.fs
```diff
@@ -1171,10 +1171,146 @@ type internal ThreadPoolWait internal (id:int) as this =
...
+        while not rwhToRemove.IsEmpty do
+            let bRemove, removeTuple = rwhToRemove.TryDequeue()
+            if bRemove then 
+                let handle, jobObject = removeTuple
+                let bExistHandleCollection, handleCollection = ThreadPoolWait.WaitHandleCollection.TryGetValue( handle )
+                if bExistHandleCollection then 
+                    let bExist, tuple = handleCollection.TryGetValue(jobObject)  
+                    if bExist then 
+                        let rwh, _, _ = tuple       
+                        rwh.Unregister(null) |> ignore
```
zhuweirong: On this code path, do you also need to remove the jobObject from the handleCollection, and remove the handleCollection from WaitHandleCollection?

src/tools/tools/process.fs
```diff
@@ -1171,10 +1171,146 @@ type internal ThreadPoolWait internal (id:int) as this =
...
+    static member TryRemove (handle:WaitHandle) = 
+        let bExist, handleCollection = ThreadPoolWait.WaitHandleCollection.TryGetValue( handle )
+        if bExist then 
+            for pair in handleCollection do 
+                let rwh, _, ticksRef = pair.Value
+                // Try to marked job as cancelled, 
+                Volatile.Write( ticksRef, DateTime.MaxValue.Ticks )
+        match handle with 
+        | :? ManualResetEvent as ev -> 
+            ev.Set() |> ignore
```
zhuweirong, line 1250: based on the code here, it seems we will set the event even when bExits = false,  why will this happen?

src/tools/tools/process.fs
```diff
@@ -1814,16 +1951,17 @@ and [<AllowNullLiteral>]
...
+//                    for tuple in ThreadPoolWait.WaitingThreads do
+//                            let (pool, _) = tuple
+//                            let continueTicks = pool.ContinueTicks 
+//                            if continueTicks >=0L then 
+//                                let cur = (PerfADateTime.UtcNowTicks())
+//                                if cur - continueTicks > ThreadPoolWithWaitHandles<_>.MaxContinuationDurationInMilliSeconds * TimeSpan.TicksPerMillisecond then 
+//                                    let elapseMs = ( cur - continueTicks )/ TimeSpan.TicksPerMillisecond
+//                                    Logger.LogF( LogLevel.Info, ( fun _ -> sprintf "Possible Deadlocks! cont() of %s has been in executionin ThreadPoolWait for more than %d ms. Possible deadlock or poor performance in implementing cont() (there should not be any blocking operation or long running operation in cont()) " 
+//                                                                                       (pool.ContinueInfo())
+//                                                                                       elapseMs ))
```
zhuweirong, line 1941: just remove the dead code?

src/tools/tools/process.fs
```diff
@@ -1171,10 +1171,146 @@ type internal ThreadPoolWait internal (id:int) as this =
...
+                    ThreadPoolWait.ToUnregister()
+                else
+                    rwhToRemove.Enqueue( handle, jobObject )
+            with 
+            | ex -> 
+                Logger.LogF( LogLevel.Error, fun _ -> sprintf "ThreadPoolWait.Callback, exception: %A" ex )
+                reraise()
+        else
+            failwith "ThreadPoolWait.Callback, timeout path reached (impossible execution route)"
+    static member ToUnregister() =
```
zhuweirong: should be private

src/tools/tools/process.fs
```diff
@@ -1171,10 +1171,146 @@ type internal ThreadPoolWait internal (id:int) as this =
...
+            ThreadPoolWait.TryRemove( pair0.Key ) |> ignore 
+        ()
+    /// Mark all functions associated with handle as cancelled, the continuation function will not execute
+    static member TryRemove (handle:WaitHandle) = 
+        let bExist, handleCollection = ThreadPoolWait.WaitHandleCollection.TryGetValue( handle )
+        if bExist then 
+            for pair in handleCollection do 
+                let rwh, _, ticksRef = pair.Value
+                // Try to marked job as cancelled, 
+                Volatile.Write( ticksRef, DateTime.MaxValue.Ticks )
```
zhuweirong: Just to call out that this is another place that we rely on 64-bit read/write being atomic.

zhuweirong:
The 2nd run failed: https://ci.appveyor.com/project/zhuweirong/prajna-rbpft/build/0.0.1.260, I'm re-running it again.

src/tools/tools/process.fs
jinlmsft: Will redesign. Use only jobObject as a single level ConcurrentDictionary, instead of two level.

src/tools/tools/process.fs
```diff
@@ -1171,10 +1171,97 @@ type internal ThreadPoolWait internal (id:int) as this =
...
+/// ThreadPoolWait.RegisterWaitForSingleObject, with capability:
+/// 1. To remove a particular wait handle. 
+/// 2. To show all wait handles queued in the system, 
+/// and 3. To show long continuation (not normal). 
+type internal ThreadPoolWait() = 
+    static let rwhToRemove = ConcurrentQueue<_>()
+    static let rwhLookup = ConcurrentDictionary<_,_>()
+    static let waitHandleCollection = ConcurrentDictionary<_,_>() 
+    static let unregisterTimer = new System.Threading.Timer( TimerCallback( ThreadPoolWait.ToUnregister ), null, 10, 10 )
+    static member WaitForHandleWithCancellation (infoFunc: unit-> string) (handle:WaitHandle) (continuation:unit->unit) (unblockHandle:ManualResetEvent) (token:CancellationToken) =
```
zhuweirong: the "token" is not used. Is it for future use?  please add a comment to avoid the confusion here?

src/tools/tools/process.fs
```diff
@@ -1171,10 +1171,97 @@ type internal ThreadPoolWait internal (id:int) as this =
...
+        token.Register( fun _ -> ThreadPoolWait.TryWakeup handle |> ignore ) |> ignore
```
zhuweirong: should we first check if the token has already received the cancellation request?

zhuweirong: (Even with the check I mentioned above) it is possible that we register the "TryWakeUp" with the token,  but a cancellation request happens before we move to the WaitForHandle (or in mid of it), the TryWakeUp is called and did nothing, and we go ahead with WaitForHandle.   The point is:  register TryWakeUp with token does not ensure it will indeed do the wakeup.  So we cannot completely rely on it.  Do we have any other measures? If we do, is this  "register" still useful?

# 

I just realized that TryWakeUp is only to signal the handle to wait, it should be fine then ...

src/tools/tools/process.fs
```diff
@@ -1171,10 +1171,97 @@ type internal ThreadPoolWait internal (id:int) as this =
...
+        let requeue = List<_>()        
+        while not rwhToRemove.IsEmpty do
+            let bRemove, jobObject = rwhToRemove.TryDequeue()
+            if bRemove then 
+                let bExist, tuple = waitHandleCollection.TryGetValue( jobObject )
+                if bExist then 
+                    let rwh, _, _, _ = tuple
+                    rwh.Unregister(null) |> ignore 
+                else
+                    requeue.Add( jobObject )
```
zhuweirong, line 1237: in what scenario that it is possible that the jobObjects exists in rwhToRemove but not in waitHandleCollection?  If happens, should this be an error that we should log and investigate?

src/tools/tools/process.fs
```diff
@@ -1171,10 +1171,97 @@ type internal ThreadPoolWait internal (id:int) as this =
...
+        for jobObject in requeue do 
+            rwhToRemove.Enqueue()
```
zhuweirong: hmm, the Enqueue does not need to take any parameter?

src/tools/tools/process.fs
```diff
@@ -1171,10 +1171,97 @@ type internal ThreadPoolWait internal (id:int) as this =
...
-            
-
+          
+/// ThreadPoolWait.RegisterWaitForSingleObject, with capability:
+/// 1. To remove a particular wait handle. 
+/// 2. To show all wait handles queued in the system, 
+/// and 3. To show long continuation (not normal). 
+type internal ThreadPoolWait() = 
+    static let rwhToRemove = ConcurrentQueue<_>()
+    static let rwhLookup = ConcurrentDictionary<_,_>()
```
zhuweirong: rwhLookup is not used anywhere.

src/tools/tools/process.fs
```diff
@@ -1171,10 +1171,97 @@ type internal ThreadPoolWait internal (id:int) as this =
...
+                else
+                    requeue.Add( jobObject )
+        for jobObject in requeue do 
+            rwhToRemove.Enqueue()    
+    static member TryTerminateAll() = 
+        for pair in waitHandleCollection do
+            let _, _, _, handle = pair.Value
+            ThreadPoolWait.TryWakeup( handle ) |> ignore 
+        ()
+    /// Mark all functions associated with handle as cancelled, the continuation function will not execute
```
zhuweirong: The comment is not accurate.  The continuation function will be executed, we are depending on the caller to cancel them properly before calling this function.  (BTW: this is the semantics I prefer).

src/tools/tools/process.fs
```diff
@@ -1171,10 +1171,97 @@ type internal ThreadPoolWait internal (id:int) as this =
...
+    static member TryWakeup (handle:WaitHandle) = 
+        match handle with 
+        | :? ManualResetEvent as ev -> 
+            ev.Set() |> ignore 
+            true
+        | _ -> 
+            false
```
zhuweirong, line 1267: Let's log here so we can identify such use cases.

PR #79, opened 2015-10-17
Alternative Distributed function executor by SemaphoreSlim.

jinlmsft:
Alternative Distributed function executor by SemaphoreSlim.

### DIFF
--- a/src/CoreLib/distributedfunction.fs
+++ b/src/CoreLib/distributedfunction.fs
@@ -206,20 +206,20 @@ type internal SchemaBinaryFormatterHelper<'T>() =
         if Utils.IsNotNull o then o :?> 'T else Unchecked.defaultof<_>
 
 /// Helper class to generate schema ID
-type internal SchemaPrajnaFormatterHelper<'T>() = 
-    /// Get a GUID that representing the coding of a type. Note that this binds to the 
-    /// fullname of the type with assembly (AssemblyQualifiedName) so if the assembly 
-    /// of the type changes, the schemaID will change. 
-    static member SchemaID() = 
+type internal SchemaPrajnaFormatterHelper<'T>() =
+    /// Get a GUID that representing the coding of a type. Note that this binds to the
+    /// fullname of the type with assembly (AssemblyQualifiedName) so if the assembly
+    /// of the type changes, the schemaID will change.
+    static member SchemaID() =
         let schemaStr = "Prajna.Tools.BinarySerializer:" + typeof<'T>.AssemblyQualifiedName
         let hash = HashStringToGuid(schemaStr)
         hash
-    static member Encoder(o:'T, ms:Stream) = 
+    static member Encoder(o:'T, ms:Stream) =
         let fmt = Prajna.Tools.BinarySerializer() :> IFormatter
         fmt.Serialize( ms, o )
-    static member Decoder(ms:Stream) = 
+    static member Decoder(ms:Stream) =
         let fmt = Prajna.Tools.BinarySerializer() :> IFormatter
-        let o = fmt.Deserialize( ms ) 
+        let o = fmt.Deserialize( ms )
         if Utils.IsNotNull o then o :?> 'T else Unchecked.defaultof<_>
 
 /// Govern the behavior of the default serialization to be used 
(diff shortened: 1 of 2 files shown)


src/CoreLib/distributedfunction.fs
```diff
@@ -343,6 +343,115 @@ type internal DistributedFunctionHolder(name:string, capacity:int, executor: (Gu
...
+    override x.ToString() = sprintf "Distributed function %s:%s/%d" name (if Utils.IsNull semaphore then "Unknown" else (capacity-semaphore.CurrentCount).ToString() ) (capacity)
+    /// Cancel all jobs related to this distributed function. 
+    member x.Cancel() = 
+        // This process is responsible for the disposing routine
+        cts.Cancel()
+    /// Can we dispose this job holder?
+    member x.CanDispose() = 
+        // Wait for all job in semaphore to exit
+        if capacityControl then 
+            semaphore.CurrentCount = capacity
```
zhuweirong, line 430: if semaphore.CurrentCount < capacity, this will return false,  then in x.CleanUp, cts.Dispose will not be called, but it should be.  

Also "CurrentCount" means the number of remaining threads that can enter the semaphore, it does not mean the # of threads that has entered the semaphore.

src/CoreLib/distributedfunction.fs
```diff
@@ -343,6 +343,115 @@ type internal DistributedFunctionHolder(name:string, capacity:int, executor: (Gu
...
+        if bCancelled then 
+            // Operation has already been cancelled. 
+            observer.OnCompleted()
+        else
+            let useTimeBudget = if timeBudget <=0 then Int32.MaxValue else timeBudget 
+            let launchTicks = DateTime.UtcNow.Ticks 
+            let chainedCTS = CancellationTokenSource.CreateLinkedTokenSource( token, cts.Token )
+            if capacityControl then 
+                let semaTask = semaphore.WaitAsync(useTimeBudget, chainedCTS.Token )
+                let wrapperExecute (taskEnter:Task<bool>) =
```
zhuweirong: does the wrapperExecute need to capture a token from chainedCTS, so it knows how to cooperatively cancel the work?

src/CoreLib/distributedfunction.fs
```diff
@@ -343,6 +343,115 @@ type internal DistributedFunctionHolder(name:string, capacity:int, executor: (Gu
...
+                            Logger.LogF( LogLevel.Warning, fun _ -> sprintf "ExecuteWithTimebudget, Functioon %s, exception while not enter, of %A" name ex )
+                            observer.OnError( ex )
+                            chainedCTS.Dispose()
+                let cancelledExecute (taskEnter:Task<bool>) = 
+                    try 
+                        chainedCTS.Dispose()
+                        semaphore.Release() |> ignore 
+                    with
+                    | ex -> 
+                        Logger.LogF( LogLevel.Warning, fun _ -> sprintf "ExecuteWithTimebudget, Functioon %s, exception when dispose of %A" name ex )
```
zhuweirong, line 410: also dispose chainedCTS here?

src/CoreLib/distributedfunction.fs
jinlmsft, line 430: The idea is when .Cancel() is called, no more new works will enter semaphore. We will wait for work to exit, before calling dispose. 

If semaphore.CurrentCount < capacity, some work may still be in execution, disposing at that time seems still pre mature.

src/CoreLib/distributedfunction.fs
jinlmsft, line 410: If an exception has filed in disposing chainedCTS.Dispose()  and release semaphore.Release(), there is no point in trying to disposing again (2nd exception?)

src/CoreLib/distributedfunction.fs
zhuweirong, line 430: I see. What's the mechanism to wait for these on-going work to exit? So one can safely call dispose?

I'm merging the pull request now to unblock you. This question can be answered/addressed later.

jinlmsft, line 430: Distributed function calls a user provided function, which is not guaranteed to behave nicely. The user provided function itself can hang forever (meaning the executor(_) never returns). 

For caller, we can return a timeout exception. 

For callee (provider), I felt that there is no easy way to prevent malfunction of the provider. 

One possibility is that for those executor which implements cooperative cancellation, the operation can be cancelled if a system time out value is reached (say 600 sec). Even such timeout behavior is tricky to implement, as a system wide timeout value is risky to implement (it is highly possible that some rate operation will indeed need to take a long time to return). 

For executor which doesn’t implements cooperative cancellation, which will be the majority of the case for the distributed function, there is probably no way out (no nice way to cancel a function once it launches). 

Jin

PR #82, opened 2015-10-20
Fix issues related to DSet.store.

jinlmsft:
Bruno alert me that DSet.store is broken on the remote cluster. This check-in fix the issue, which is the (Echo,DSet) not properly formatted with Job information, and causing DSet.store to report
error during execution.


zhuweirong:
Hi, this commit includes a lot more stuff (such as built-in function etc) than what the commit note suggests. Is this intentional? If so, I'd recommend update the commit note to reflect the intention.

jinlmsft:
I just started to work on the built-in function when I switched gear to fix Bruno’s issue. I should start a new branch for the fix. (My bad to reuse the branch of Distributed Function). 

Most of the fix is geared towards the fix. So the commit note is about right. The only exception is the change in the builtinfunction.fs. Currently, it is just a place holder, and the implementation hasn’t completed yet. 

Jin

PR #91, opened 2015-10-27
Logging API updates

zhuweirong:
1. The LoggerProvider is more like a storage layer for logs. It should not have APIs regarding to job id, which is a logical concept for Prajna. So move JobID related API to logger class
2. Whether a log level is enabled or not, should be the logger provider's decision. It was often known only at runtime via external configurations.


src/tools/tools/Logger.fs
```diff
@@ -363,56 +367,23 @@ type Logger internal ()=
...
+    static member DefaultLogLevel() =
+        Logger.LoggerProvider.MaxLogLevelEnabled(Logger.DefaultLogId)
+
+    static member DefaultLogLevel(logId) =
+        Logger.LoggerProvider.MaxLogLevelEnabled(logId)
+
     /// Log "message" if logLevel <= Logger.DefaultLogLevel                                                                    
     static member inline Log(logLevel : LogLevel, message : string) =
-        if logLevel <= Logger.DefaultLogLevel then
+        if Logger.LoggerProvider.IsEnabled(Logger.DefaultLogId, logLevel) then
```
jinlmsft: My understanding is that with this change, Log message that are suppressed in the past (e.g., WildVerbose) will go through one more function call of IsEnabled(). Will that cause performance issue? Can you check with Sanjeev on the performance with related to sort before and after the change?

Thanks,

zhuweirong:
It does not only add a function call but a virtual function call. In this case, a very good JIT may still be able to inline it (since the call target is always the same) but I'm not sure whether CLR JIT would do it or not.  This is the cost of abstracting LogProvider so to be able to attach different ones. Whether a logging level (for a log id) is allowed or not is normally decided via an external configuration files. For example, take a look at NLog, which allows specify level as rules in configuration file (https://github.com/nlog/NLog/wiki/Configuration-file#rules). 

Since there're performance concerns, I will delay this change until you can discuss this in person. Meanwhile, I will see if Sanjeev can help run some perf test using sort benchmark.

PR #101, opened 2015-11-24
Fix the NetworkConnects initialization issue, by Sanjeev

leizhangcn:


jinlmsft:
Thanks for the explanation. I will check this in. Although it might be a better idea to properly initialize, e.g., calling Environment.Init()

PR #105, opened 2015-12-02
Workaround for the issue that Start-Client.ps1 fail to launch the client

zhuweirong:
Workaround for the issue that Start-Client.ps1 fail to launch the client when no active user session on remote machine.  It seems adding some tiny sleep after Invoke-WmiMethod alleviates the issue. More investigations are needed.

### DIFF
--- a/scripts/WindowsDomainCluster/Start-Client.ps1
+++ b/scripts/WindowsDomainCluster/Start-Client.ps1
@@ -83,6 +83,8 @@ $Sb = {
 
     $result = Invoke-WmiMethod -path win32_process -name create -argumentlist @($Cmd, $ClientLocation)
 
+    Start-Sleep -Seconds 1
+
     $Procs = Get-Process -Name $ClientNameWithoutExtension -ErrorAction SilentlyContinue
     if ($Procs)
     {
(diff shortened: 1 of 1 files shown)


zhuweirong:
No. This work-around still needs to happen within each session that connects to a single machine. So the sleep needs to happen in the script block inside Start-Client.ps1. 

However, a future work-item is to make Deploy-Clients.ps1 and Start-Clients.ps1 invoke Start-Client.ps1 in parallel for multiple machines.

jinlmsft:
I assume that a single session is Deploy-Clients.ps1 and Start-Clients.ps1, not Start-Client.ps1, is that true?

Indeed, launch Start-Client.ps1 in parallel may solve the problem. And I am fine with that solution. In such an instance though, multiple machine may launch robocopy of the source. If the concurrent connection is too many, Windows sometime will complain.

Best,
Jin

zhuweirong:
No. There is a session between the host machine to each remote machine. If you have N machines, there are N independent sessions.

jinlmsft:
Yes, you are right (as always). I confused the local session with the remote connected session to launch the program. Please merge the change.

Best,
Jin

zhuweirong:
To clarify, the session here refers to PSSession that is used to issue command to a remote machine. 

Also, the script currently does not use Robocopy, but use Copy-Item cmdlet. If there're too many machines, we should not blindly launch parallel jobs for each individual machine. Otherwise, we may suffer the problem you mentioned. We can set some threshold on max parallelism.

PR #108, opened 2015-12-03
Fixing serialization of arrays of element type with size < 4

BrunoBozza:
Also adding MemoryStreamB to all serialization tests


zhuweirong:
The CI build failed, please take a look.

zhuweirong:
Please also remove "Ignore" for a repro test at https://github.com/MSRCCS/Prajna/blob/master/tests/CoreLib/DSet.fs#L1643, or if you feel this test is redundant,  please remove the test completely.

PR #120, opened 2015-12-15
Non-Serializable, circular C# closures, perf, tests

BrunoBozza:
- Adding support for classes not marked as [Serializable] for Portable Class Library compatibility, with test
- Adding support for circular C# closures, which use circular IObjectReferences (BinaryFormatter will error-out on these, we can't fix with SurrogateSelector unfortunately)
- Specialized code for common zero-based, 1D arrays, for common-case speed
- New perf tests comparing BinarySerializer and PrajnaSerializer
- Modifying tests to run with both MemoryStreamB and standard .Net MemoryStream

### DIFF
--- a/Prajna.sln
+++ b/Prajna.sln
@@ -150,6 +150,8 @@ Project("{8BC9CEB8-8B4A-11D0-8D11-00A0C91BC942}") = "qsort", "samples\SortBenchm
 EndProject
 Project("{2150E333-8FDC-42A3-9474-1A3956D46DE8}") = "SortBenchmark", "SortBenchmark", "{9DDC2A59-C4D8-4DDE-B82E-A37DF325F109}"
 EndProject
+Project("{FAE04EC0-301F-11D3-BF4B-00C04F79EFBC}") = "CSharpTests", "tests\CSharpTests\CSharpTests.csproj", "{6797EE77-90E4-4719-BC81-F3FA2840CA9A}"
+EndProject
 Global
 	GlobalSection(SolutionConfigurationPlatforms) = preSolution
 		Debug|Any CPU = Debug|Any CPU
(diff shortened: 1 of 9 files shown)


BrunoBozza:
Had forgotten to add a couple of files. Updating.

Prajna.sln
```diff
@@ -572,6 +574,22 @@ Global
 		{E95D819A-B8E3-4F94-B7EE-300E31AB0181}.Release|Win32.Build.0 = Release|Win32
 		{E95D819A-B8E3-4F94-B7EE-300E31AB0181}.Release|x64.ActiveCfg = Release|x64
 		{E95D819A-B8E3-4F94-B7EE-300E31AB0181}.Release|x64.Build.0 = Release|x64
+		{6797EE77-90E4-4719-BC81-F3FA2840CA9A}.Debug|Any CPU.ActiveCfg = Debug|Any CPU
```
zhuweirong: To be consistent with all other projects within this sln, it should only contain x64.

tests/tools/tools/serialize.fs
```diff
@@ -291,4 +297,95 @@ module SerializationTests =
...
+        Assert.IsTrue(prajnaTime.TotalMilliseconds < binaryTime.TotalMilliseconds)
+        
+    let createRandomJaggedMatrix rows cols = 
+        let r = new Random()
+        Array.init rows (fun _ -> Array.init cols (fun _ -> r.NextDouble() |> float))
+
+    let baseSize = 5000
     
+    [<Test>]
+    [<Category("Performance")>]
```
zhuweirong, line 352: These tests will not be ran as part of the build since they are excluded. But they will show up in VS test explorer, which probably does not have the functionality to exclude based on NUnit category.  If that's the case, we need to add some note to the wiki (https://github.com/MSRCCS/Prajna/wiki/Contribute, at the "Visual Studio" section) to call it out.  So when ppl do a "run all" in test explorer, they won't be surprised.

Prajna.sln
BrunoBozza: Made consistent now.

tests/tools/tools/serialize.fs
BrunoBozza, line 352: Added note to wiki.

PR #134, opened 2016-01-12
Fix issue regarding potential race condition at connection initialization.

sanjeevm0:
1. Add *.opendb to gitignore
2. Upgrade compiler to VS2015
3. Attempt to fix issue found regarding message being sent prior to initial connection
   being verified.  This can cause incorrect message parsing.
4. Add aligned array - can be used when interop with native code.
5. Fix one issue with mutable variable not being captured by closure (VS2012 does not compile)

### DIFF
--- a/src/CoreLib/network.fs
+++ b/src/CoreLib/network.fs
@@ -747,7 +747,7 @@ type [<AllowNullLiteral>] NetworkCommandQueue internal () as x =
 
     member private x.AsyncSendBufWSizeTryCatch(buf : byte[], bufferOffset : int, bufferSize : int) =
         try
-            xgBuf.AsyncSendBufWithSize(buf, bufferOffset, bufferSize)
+            xgBuf.AsyncSendBufWithSize(None, null, buf, bufferOffset, bufferSize)
         with e ->
             x.MarkFail()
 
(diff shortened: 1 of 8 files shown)


src/CoreLib/network.fs
```diff
@@ -286,6 +286,10 @@ type [<AllowNullLiteral>] NetworkCommandQueue internal () as x =
 
     let eVerified = new ManualResetEventSlim(false)
 
+    let eAsyncSend = new ManualResetEventSlim(true)
```
jinlmsft: Do we expect to wait on eAsyncSend using WaitForSingleObject (or WaitForHandle)? 
The current code may have issues if there are multiple AsyncSends outstanding. Say we have AsyncSend1, AsyncSend2, and x.StartConnection(), AsyncSend1 proceeds, while AsyncSend2 and StartConnection gets blocked on eAsyncSend.Wait(), when AsyncSend1 finishes, both StartConnection  and AsyncSend2 can proceed at the same time. Will this causes an issue?

It looks to me that it might be simpler to convert eAsyncSend to an object, and just lock the object using lock(eAsyncSend). We will avoid an object to dispose. lock(o) uses Monitor.Enter, which uses SpinLock and keyed event internally, and may lead to better efficiency.

sanjeevm0:
Will create new one with new title.

PR #147, opened 2016-01-30
Thread pool

jinlmsft:
Rewrite Threadpool. Add additional support to trace bugs in Mono.

### DIFF
--- a/build.fsx
+++ b/build.fsx
@@ -422,6 +422,7 @@ let runTests (target:string) =
             timer.Start()
         try
             let pattern = String.Format(testAssemblies, target)
+            traceImportant (sprintf "Running NUnit with parameter %s" pattern)
             !! pattern
             |> NUnit (fun p ->
                 { p with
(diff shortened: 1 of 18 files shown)


dsyme:
Any idea what's up with the Travis build? 

Best
don

msrccsbuild:
It seems that paket breaks, which causes Travis build to fail. More specifically, Paket fails to pull down the following NuGet packages FSharpVSPowerTools.Core 2.3.0, and Octokit 0.18.0, which causes Travis build to fails.

Interestingly, the package gets successfully pulled during windows build, but fails in Linux build (I have verified that the pull fails in a Linux machine).

I assume that this should affect other Paket/Scaffold users, and are waiting for a fix from Paket/FSharpVSPoweTools/Octokit. I don't know who to report this issue.

Best,
Jin

dsyme:
I suppose report it to https://github.com/fsprojects/Paket if you think it's paket.

jinlmsft:
I am not sure what is the root cause. It can be either: 1) Paket, 2) FSharpVSPoweTools/Octokit. 

Jin

PR #161, opened 2016-02-25
Attempt fix for hang issue during cleanup.

sanjeevm0:
Set inProcessing flag to false prior to calling closeAction


msrccsbuild:
Sanjeev,

I may have concern on whether F# properly honor the VolatileField attribute, and from the disassembly I see, I would argue that F# does not properly use VolatileField. I would recommend that you use Volatile.Read and Volatile.Write to expressly read/write all volatile variable. Otherwise, memory reordering may cause tricky bugs in the code.

Best,
Jin

PR #168, opened 2016-03-03
Increase timeout in case of long network upload of assembly / dependency

sanjeevm0:


jinlmsft:
Discussed with Sanjeev, I will check-in a fix that covers this one.

